PR #55, opened 2026-06-30
feat(bench): synthetic hallucination-reduction benchmark for the guard

inth3shadows:
## Why

RunEcho's headline claim — *catches invented symbols / reduces hallucinated code changes* — has never been measured. No `bench/` or `eval/` existed. This adds the first number behind it.

## What

Scores `guard.Run` as a **binary classifier** over symbol references: a reference to an absent symbol is a hallucination the guard should flag; a reference to a real symbol should pass. Reports **catch-rate (recall)**, **false-positive rate** (the adoption-killer), precision, and F1 — overall, **per-language**, and **per-hallucination-type** (typo / plausible-affix / wrong-case / invented / renamed-away).

The per-language/per-type cut is the point: it makes parser-fidelity work **measurable** instead of taken on faith.

## Design

- **Deterministic by construction** — seeded perturbation of declared symbol pools; same seed → same corpus. A determinism tool gets a deterministic benchmark.
- **Isolates the classifier** — known-set is supplied directly, not IR-derived, so IR extraction accuracy (a separate claim) doesn't contaminate the score.
- **Respects per-language scope** — Go pool symbols are Exported, because the IR indexes only exported Go symbols and the guard skips unexported refs by design (`internal/guard/extract.go`). An unexported corpus would measure a non-feature. (This was caught and fixed while building: the first run showed Go at 20% catch-rate for exactly this reason.)

## Result & honest limits

Within the guard's call-position scope the classifier is exact — **100% recall / 0% false-positives** — and stays that way. Because corpus + guard are both deterministic, the test **gates that baseline as a non-flaky regression tripwire**: a parser/guard change that drops recall or raises false-positives fails the build with the exact delta.

A green run does **not** mean RunEcho catches all real-world hallucinations — synthetic perturbation saturates inside call-position scope and can't probe the hard cases (qualified `obj.Method`, unexported Go, non-call positions). The true number against an **observed LLM error distribution** comes from a captured-LLM corpus (Phase 2, scoped separately).

## Verification

`gofmt` clean, `go vet ./bench/` clean, `go test ./...` green (new `bench` package included). Zero new dependencies.

Run it: `go test ./bench/ -run TestScorecard -v`


inth3shadows:
Added a second commit: **captured-LLM corpus loader (Phase 2 scaffold)** — `bench/captured.go` + tests + `bench/captured/corpus.sample.json`.

It loads hand-labeled real-world hallucinations from frozen JSON fixtures and replays them through the *same* guard classifier, but reports **counts stratified by provenance** (observed-from-transcripts vs elicited-offline, never pooled) with an observed-majority floor. Strict load-time validation rejects self-inconsistent fixtures. Labels are established by hand against repo source — **not** by RunEcho (no grading the tool with itself).

The 6 sample fixtures are illustrative only; the real ~30-case corpus is harvested in a working session. The sample already shows the expected honest miss — a qualified `obj.Method` call the guard skips by design — proving the harness surfaces blind spots rather than hiding them.

Both the synthetic tripwire and this loader are source-agnostic infra with zero new deps; happy to split into a stacked PR if you'd rather keep this one purely synthetic.

PR #57, opened 2026-06-30
feat(guard): extract non-call references — const refs + type annotations (closes #56)

### LINKED ISSUE #56: guard: extract refs in non-call positions (type annotations, const/value refs) — caught by the captured benchmark
## Evidence
The new captured-LLM benchmark (`bench/`, see `bench/FINDINGS.md`) measured the guard against 15 hand-verified real hallucinations mined from session transcripts. Result: **caught 1 of 9**, and the position breakdown shows why —

```
by reference position:
  bare-call         caught 1/1   <- only position the guard covers
  const-ref         caught 0/1   <- TASTING_ROOM_KIND[t]
  type-ref          caught 0/2   <- ctx: RouteContext<...>
  qualified-attr    caught 0/3   (out of scope by design - needs type resolution)
  qualified-method  caught 0/1   (out of scope by design)
```

## The actionable gap (this issue)
`internal/guard/extract.go::ExtractRefs` only treats an identifier as a reference when it appears **before `(`** (call position). Real hallucinations also occur as:
- **bare type annotations** - `ctx: RouteContext<...>` (TS), where `RouteContext` is a bare, unqualified name absent from the IR.
- **bare constant/value references** - `TASTING_ROOM_KIND[t]` (Py), a name used as a subscript target.

These are **unqualified** identifiers - the guard could validate them against the IR known-set with **no type inference**, staying within RunEcho's deterministic-structural design. Extending extraction to these positions would have caught ~3 of the 8 misses above.

## Explicitly NOT in this issue
Qualified references (`series.pearson_corr`, `df.groupby`, `tree.Root`) require resolving the receiver's type - semantic analysis RunEcho deliberately avoids. Those should be documented as a known boundary in the README, not chased here.

## Acceptance
- Add the captured cases (`obs-ts-001/002`, `obs-py-001`) as the regression target.
- Extend `ExtractRefs` (or add a sibling extractor) to surface bare identifiers in type-annotation and value positions, language-aware, without raising false-positives on the synthetic corpus (which must stay 100%/0%).
- Re-run `go test ./bench` - the captured catch-rate should move from 1/9 upward; FINDINGS.md numbers update.

Sub-pattern worth a follow-up: several misses were **dropped imports** (`ULID`, `TASTING_ROOM_KIND`) - cross-checking refs against the file's own import set is another cheap structural signal.

inth3shadows:
Closes #56.

## What
Extends the guard beyond call-position references to two non-call positions that the [captured-LLM benchmark](bench/FINDINGS.md) showed real models hallucinate in:
- **SCREAMING_SNAKE constant refs** (Python) — e.g. a dropped-import constant `TASTING_ROOM_KIND[t]`.
- **PascalCase type annotations** (TS/JS) — e.g. `ctx: RouteContext<...>`.

## Measured effect
| | before | after |
|---|---|---|
| captured catch-rate | 1/9 | **4/9** |
| captured false positives | 0/6 | **0/6** |
| synthetic tripwire | 100%/0% | **100%/0%** |

The position breakdown: `const-ref` 0/1→1/1, `type-ref` 0/2→2/2. Remaining misses are all **qualified** positions (`obj.method`, `pkg.x`) — out of scope by design (they need receiver-type resolution, which RunEcho deliberately avoids).

## Precision (the false-positive rate is the adoption-killer)
Kept deliberately high-signal:
- const refs require an underscore-joined SCREAMING case (ordinary locals don't use it); skip qualified attrs and assignment/annotation targets.
- type refs require a lowercase-led `param: Type` shape; skip TS primitives + utility types, single-char generics (`T`/`K`/`V`), and qualified types.
- To avoid false-positiving on **local** definitions, `ExtractDefs` now also captures Python `class` + SCREAMING_SNAKE const defs and TS `interface`/`type`/`enum`/`class`; `isImportLine` skips import/re-export binding lines (but not `export function|const|...`, whose param types are real references).

## Tests
Positive **and** negative coverage for both positions (the negatives pin that the FP surface stayed flat), `ExtractDefs` type-decl capture, and an end-to-end `Run` test proving a locally-defined type resolves while an undefined one is flagged. Full guard suite + `go test ./...` green.


inth3shadows:
Superseded by a clean, rebased branch (the original was branched pre-squash-merge of #55 and carried already-merged commits → conflict). Reopening the same change on top of current master.

Issue #204, opened 2026-07-23
investigate: can RunEcho check design tokens / style values? (split the ask before building)


inth3shadows:
## Origin

Raised by a frontend user in a Slack thread (2026-07-22), working in an old,
large codebase with an agent-assisted design workflow. Their problem split in two:

1. **"This code is so old it's hard to find where it lives."** Going from "some
   component sets this" to the actual file. → RunEcho already helps here; this is
   what `locate` is for.
2. **"It fixes 70% and I hunt the missing 30%."** The remainder is margins,
   paddings and colors living in `className` strings, CSS, and styled-props.
   → **RunEcho has no representation of a style value and will not find these.**

A public commitment was made to investigate (2), gated on whether it is a real
pain for more than one person. This issue is that investigation — **not** an
agreement to build it.

## Why it does not work today

Verified against source, not assumed. The IR's per-file shape is
`internal/parser/parser.go` → `FileStructure`, and it is entirely name lists:

```go
type FileStructure struct {
	Imports   []string
	Functions []string
	Classes   []string
	Exports   []string
	WildcardReexports []string
	SymbolHashes map[string]string  // kind:name -> body hash
	SymbolLines  map[string]int     // kind:name -> start line
}
```

There is nowhere for a value to live. JSX/TSX files are already parsed by the JS
parser, so the `className` strings are physically read and then discarded —
extracting them is a schema change plus new parsers (there is no CSS parser at
all; `internal/ir/generator.go:128` registers JS, Go, Python, shell, Rust, Ruby).

## The split this investigation has to make first

Do not treat "index style values" as one feature. It is two, and they land in
opposite lanes of the competitive map:

### (a) Value lookup — "where is this padding set?"

**This is the fight to decline.** Symbol existence is *decidable*; a style value
is not. `procesData()` is undefined and provably wrong. `padding: 13px` is not
wrong — it is just a number, and no parser can tell you it was the wrong one. So
the guard can never gate on it, which means this could only ever ship as a search
feature.

As a search feature it competes with `ripgrep`, which is faster, already installed,
and does not need an index. This was said plainly in the originating thread and it
should not be quietly walked back here: **that half is closer to a ripgrep or
stylelint sweep than to anything RunEcho does.**

See also #175 — a previous "index one more thing" feature that shipped, moved the
corpus by zero, and was reverted. The premise there was also plausible.

### (b) Design-token existence — "is `text-brand-primry` a real token?"

[...]

Issue #203, opened 2026-07-23
competitive: add a raw repo-search sweep to the quarterly moat re-verification


inth3shadows:
## Problem

The quarterly moat re-verification (`~/.local/bin/runecho-moat-reverify`, first
scheduled fire 2026-10-22) surveys product docs, changelogs, PyPI pages and web
search. On the 2026-07-23 pass it produced a confident **"no new entrants found
this pass"** verdict.

That verdict was wrong. [`Akunimal/Anti-Hallucination-MCP`](https://github.com/Akunimal/Anti-Hallucination-MCP)
— created 2026-05-31, pushed 2026-07-20, checks the same thing RunEcho checks with
the same deterministic technique family — was found minutes later by a single raw
GitHub repo search.

**The method has a structural blind spot: it can only see competitors that market
themselves.** A 0-star repo with no blog post, no release notes and no directory
listing is invisible to every source the routine consults — and that is precisely
the profile of the most likely rival in this niche, because it is RunEcho's own
profile.

## What to change

Add a repo-search sweep to the routine, and make it diff-based rather than
judgement-based:

```
gh api "search/repositories?q=hallucination+symbol+claude-code&sort=updated" \
  --jq '.items[] | "\(.stargazers_count)★ \(.full_name) pushed=\(.pushed_at[0:10])"'
```

- Run several query variants, not one (`pretooluse hook guard symbol`,
  `undefined symbol agent guard mcp`, and the query above).
- **Persist the returned repo set** and diff it against the previous pass. A new
  full-name appearing is the trigger to run the four-conjunct test; an existing
  entry's `pushed_at` moving is the trigger to re-read its hook wiring.
- For any candidate that checks symbol existence, **read the installed hook
  configuration** (`PreToolUse` vs `PostToolUse`, deny vs stderr) rather than the
  README's adjectives. "Guard", "blocks" and "prevents" are used loosely in this
  space; the event name is not. This is the specific check that would catch
  Anti-Hallucination-MCP flipping its hook — which is currently the cheapest path
  to falsifying the moat claim, and would ship silently since that repo has no
  release-notes habit.

## Why it matters

The moat claim is now held by **conjunction, not absence** — GateGuard owns the
timing half, Anti-Hallucination-MCP owns the substance half. A routine that reports
"nobody is nearby" when two tools are each one step away is worse than no routine,
because it produces false confidence on a schedule.

The procedural fix is already written into `docs/competitive-landscape.md`
[...]

Issue #202, opened 2026-07-23
discovery: the pre-write-gate lane is unclaimed, not contested — lead with determinism


inth3shadows:
## Opportunity

The 2026-07-23 competitive survey turned up a finding that matters more than any
of the technical gaps it found:

**The pre-write-gate lane contains exactly two repositories, and both have 0 stars.**

```
gh api "search/repositories?q=hallucination+symbol+claude-code&sort=updated" \
  --jq '.items[] | "\(.stargazers_count)★ \(.full_name)"'
# 0★ inth3shadows/runecho
# 0★ Akunimal/Anti-Hallucination-MCP
```

That lane is **unclaimed, not contested**. Meanwhile the adjacent code-intelligence
lane — which RunEcho deliberately does not compete in — has tools at 47.4k, 42k and
25.2k stars. The gap is not capability. RunEcho holds all four conjuncts of the
pre-write test and the nearest rival holds two. The gap is that nobody can find it.

Supporting evidence: the competitive survey itself only found the rival via a raw
GitHub repo search. Product docs, changelogs and web search — three prior passes'
worth — never surfaced a project that had existed for two months. If the survey
method couldn't see them, buyers can't see us.

## The line that actually lands

A real reader (via a Slack thread, 2026-07-22) highlighted exactly one sentence
from the README:

> **The same code produces the same answer.**

Not the hallucination framing. Not the symbol index. **Determinism.** That is the
sentence that got a "you were looking for something like this" reply, and it is
currently buried at the bottom of the opening section, below a paragraph that leads
with a miss rate.

The 4-of-9 honesty is a genuine asset and should stay — but it is the *second*
thing a reader should hit, not the first. Right now the README argues against
itself before it has made its case.

## What to do

- **README first screen.** Lead with determinism and pre-write timing; keep the
  4-of-9 scope paragraph, move it below the value proposition. It reads as
  credibility once someone wants the thing, and as a warning label before they do.
- **Repo metadata.** Set GitHub topics so the search above and its neighbours
  actually return this repo — currently it is found by that query only because the
  description happens to contain the words. Compare against the rival's topics:
  `ai-safety`, `anti-hallucination`, `claude-code`, `codex`, `mcp`, `mcp-server`,
  `multi-language`, `symbol-checking`.
- **Directory presence.** MCP server directories, `awesome-claude-code`, and the
  hooks/guardrails round-ups that already enumerate this category. The rival appears
  in none of them either — this is first-mover, not catch-up.
[...]

Issue #201, opened 2026-07-23
bench: measure and publish context-token cost against the code-intelligence field


inth3shadows:
## Opportunity

The entire code-intelligence category — CodeGraph (47.4k★), GitNexus (42k★),
Serena (25.2k★) and a dozen others — is sold on **saving context**. An independent
benchmark across 8 repos and 4 models found the opposite result:

| Tool | Tokens vs no MCP at all |
|---|---|
| CodeGraph | +22% to +27% |
| codebase-memory-mcp | up to +66% |
| Serena | up to +93% |

Source: <https://dev.to/samchon/i-made-ts-compiler-graph-mcp-10x-fewer-tokens-in-claude-code-1aea>
(measured on the `zod` repo among others). The author's diagnosis: tool-call count
falls, but the tools either hand back whole source bodies or bury capability the
agent can't find, so it reverts to grep — and the thing you pay for never drops.

**RunEcho's guard is structurally exempt from this failure mode.** It is a
`PreToolUse` hook, not a tool the agent decides to call, so it costs zero context
tokens unless it actually blocks. That is currently an *assertion in the README*
("~0 tokens of your context window"), not a measured claim — and this project's
whole credibility posture is that measured beats asserted.

Nobody in the field has claimed "zero-token verification" with data behind it.

## What to build

A reproducible token-cost measurement, in the same spirit as `bench/FINDINGS.md`:

- Measure **actual context tokens consumed** across a fixed task set, comparing:
  baseline (no MCP), RunEcho guard only, RunEcho MCP tools, and at least one graph
  MCP available locally (`codegraph` is already installed and indexed on this box).
- Report the guard's zero separately from the MCP server's non-zero. The guard is
  free; `structure`/`locate` calls are not, and conflating them would be exactly
  the kind of overclaim the bench exists to prevent.
- State the methodology and its limits as plainly as `bench/FINDINGS.md` states the
  4-of-9 caveat.

## Honest risks to write into the findings

- The cited +22/+66/+93% figures are **one author, one methodology, and that author
  ships a competing MCP.** Cite the direction; do not inherit the magnitudes.
- RunEcho's 0 is a *structural property of being a hook*, not an experimental
  result. If the bench reports it as measured, the bench is lying. Measure what the
  MCP server costs and let the hook's zero be the architectural argument it is.
- A block emits tokens. Frequency of blocks × message size is the real cost and
  should be reported, not rounded to zero.

## Relationship to #173

[...]

PR #205, opened 2026-07-23
docs: correct the competitive survey, and lead the README with what RunEcho does

inth3shadows:
## What

Corrects the 2026-07-23 quarterly moat re-verification, which concluded **"no new
entrants found this pass."** That was wrong.

[`Akunimal/Anti-Hallucination-MCP`](https://github.com/Akunimal/Anti-Hallucination-MCP)
— created 2026-05-31, pushed 2026-07-20, 0★ — checks the same thing RunEcho
checks, with the same deterministic technique family: ast-grep for JS/TS/TSX,
regex fallback for ~15 more languages, confidence scoring, fuzzy typo matching.
It is the closest miss on record.

## It does not falsify the claim

| Conjunct | Verdict |
|---|---|
| 1 · Pre-write | ❌ its installed hook is `PostToolUse` |
| 2 · Blocking | ❌ warns to stderr after the write has landed |
| 3 · Symbol-existence aware | ✅ |
| 4 · Deterministic / model-free | ✅ |

Two further concessions it makes that RunEcho does not: a persisted
`.wisdom/symbols.json` registry that can lag the tree (mitigated by an opt-in
watcher, not by re-parsing), and an `ANTIHALL_AUTOFIX=1` mode that rewrites the
identifier in place — a repair posture, not a gate.

## But the shape of the field changed, and the doc now says so

- **GateGuard is the nearest neighbour on _timing_** — pre-write and blocking, but
  never asks whether a symbol exists.
- **Anti-Hallucination-MCP is the nearest neighbour on _substance_** —
  deterministic symbol-existence checking, but only after the write lands.

Each holds exactly the half the other lacks. The moat is now held by
**conjunction, not by absence**, which is a materially weaker position than the
previous survey described, and it is stated that way rather than smoothed over.
Its falsification path is roughly one line of JSON — change the event the setup
script writes and return a deny instead of a warning — so it is recorded as the
cheapest path on the watch list, with an instruction to check the hook wiring
every pass rather than the changelog (0★ repo, no release-notes habit, it would
ship silently).

## Why the survey missed it

Prior passes searched product docs, changelogs, PyPI and the web. None of those
can see a repo with no marketing surface — which is the exact profile of the
likeliest rival, and RunEcho's own profile. Added to the re-verification
procedure:

```
gh api "search/repositories?q=hallucination+symbol+claude-code&sort=updated" \
  --jq '.items[] | "\(.stargazers_count)★ \(.full_name) pushed=\(.pushed_at[0:10])"'
```

As of this pass that returns exactly two repositories: RunEcho and
Anti-Hallucination-MCP. Also added: for any candidate that checks symbol
existence, read the hook wiring (`PreToolUse` vs `PostToolUse`, deny vs stderr)
rather than the README's adjectives — "guard", "blocks" and "prevents" are used
loosely in this space; the event name is not.

## Also in this PR

`README.md` omitted Rust and Ruby from its parsed-languages line. Both shipped in
#194/#195 and are registered at `internal/ir/generator.go:128`; the line
undercounted the supported set by two.

## Verification

- Five-date-stamp invariant re-checked: `grep -ro '2026-07-23' docs/competitive-landscape.md docs/runecho-vs-field.html | wc -l` → **5**. The new procedural note is deliberately undated so it does not move with the survey stamps.
- Rival behaviour read from its README and hook configuration directly, plus GitHub API metadata — not from a summary.
- Docs only; no code paths touched.

## Follow-ups filed

#200 (did-you-mean on block) · #201 (context-token bench) · #202 (discovery) ·
#203 (repo-search sweep in the quarterly routine) · #204 (design-token
investigation), plus a mechanism comment on #174.


inth3shadows:
## Added: #202 discovery work

Third commit reorders the README opening. **Nothing is softened** — the 4-of-9 scope paragraph is unchanged and still on the first screen, just below the value proposition instead of above it.

Rationale is in the commit message. Short version: a real reader highlighted exactly one sentence out of the whole README — *"The same code produces the same answer"* — and it was buried in the fourth paragraph, below a paragraph that leads with a miss rate.

Also corrected the guard's posture from "blocked" to "stops the write and asks you first", matching the shipped ask-posture hook.

### Repo metadata — already applied, outside this diff

These are live on the repo now and are not reviewable in the PR:

- **Description** rewritten to lead with the pre-write guard rather than the hedge, keeping the keywords that make the repo findable by `hallucination symbol claude-code` (337 chars).
- **Topics** 16 → 20: added `anti-hallucination`, `symbol-checking`, `ai-safety`, `pretooluse`.

Revert either with `gh repo edit` if you disagree.

### Still open under #202

Directory listings — MCP server directories and `awesome-claude-code`. Neither RunEcho nor the one other tool in this lane appears in any of them, so it is first-mover rather than catch-up.

Issue #227, opened 2026-07-24
bench: hook-level checks have no replayable fixtures — the published catch-rate covers 1 check of 6


inth3shadows:
**Category: best practice — the pitch is "measured, not asserted", and five of six checks are unmeasured by the published harness.**

## The gap

The guard ships six checks. Five of them — `dangling-refs`, `dropped-import`, `duplicate-symbol`, `file-scope`, `contract` — need edit old-vs-new text plus an enrolled snapshot store, and are reachable **only** through the hook entry point.

Every published instrument invokes the guard in-process through `guard.Run`:

- `internal/guard/corpus_test.go` — the 47-case FP/FN regression gate
- `bench/score.go:120` — `guardFlags`, behind the synthetic scorecard and the captured N=15 corpus
- `cmd/runecho-guard/hookmode_test.go:107` — `runHook` calls `runHookMode(...)` in-process

`bench/FINDINGS.md` already says it outright for one of them: *"it is a hook-level check … so it is not exercised by this benchmark's `guard.Run` path — its validation is the unit suite."* The same is true of the other four.

So the catch-rate RunEcho publishes describes **one check of six**.

## Verified

```
$ grep -ilE "dangling|duplicate|filescope|contract|dropped.?import" internal/guard/testdata/corpus/*.json
NONE
```

Those checks' cases exist — as ~53 Go test functions across `cmd/runecho-guard/{dangling,duplicate,filescope,contract}_test.go` — but as code, not as replayable data. A harness cannot consume them.

## Scope

Convert the hook-level checks' cases into replayable JSON fixtures alongside the existing 47, carrying whatever the hook entry point needs (old/new text, the env flag that gates the check, and an enrolled snapshot fixture) rather than only the `known`/`infile`/`edit`/`expect` shape `guard.Run` takes today.

**Explicitly not "grow the corpus."** No new cases are needed — the cases exist. This is a data-shape and harness-reach problem.

## Why it matters beyond the number

This is the prerequisite for two things that are otherwise blocked:

1. **A `selftest` that replays the corpus through the INSTALLED binary** over its real stdio interface. That was proposed and rejected this round precisely because the corpus can't reach five of six checks — it would have printed 1/6 real coverage and 5/6 vacuous, dressed as a fixed instrument.
2. **#173** (public catch-rate harness others can run against their own agent). A benchmark-as-standard that silently covers one check of six is not a standard.

## Success

[...]

PR #246, opened 2026-07-28
test: replayable hook fixtures for the file-scope check (#227, phase 4)

inth3shadows:
Phase 4 of #227, after #238 (harness + duplicate-symbol), #242 (dangling-refs) and #245 (dropped-import). Covers the file-scope resolution check — the "real symbol, wrong scope" class that `bench/FINDINGS.md`'s corpus mining found to be the dominant real-world failure, and the one the additive check structurally cannot see.

## The enrollment inverts phase 3's

The symbol is enrolled in **another** file, so it is known repo-wide — which satisfies the check's firewall *and* keeps the always-on additive check silent, since that one fires only on names ABSENT from the known set. It is simultaneously unresolvable in the edited file. So an ask can only have come from file-scope.

## A harness gap this phase had to close first

Every fixture was previously replayed as a `Write`, under which `hookAddedLines` makes the added-lines set the **entire new file**. That makes any whole-file claim untestable: an abstain marker sitting in both `old` and `new` also lands in the added text, so `abstainsFileScope(addedLines)` alone triggers and the fixture would pass even if the pre-edit file read were deleted outright. The first draft of these fixtures had exactly that defect.

`tool`/`edit_old`/`edit_new` add a hunk-scoped `Edit` replay. The two abstain true-negatives now keep their marker **outside** the edited hunk, so only the on-disk file carries it, and each is paired with an identical-hunk control against a marker-free file that **does** ask — the difference in outcome is then attributable to the file text alone.

**Verified by mutation**, not by assertion: deleting the `abstainsFileScope(wholeFile) ||` arm of `internal/guard/filescope.go` fails exactly `python-star-import-abstains-clean` and `python-dynamic-binding-abstains-clean`, and nothing else.

## The 8 fixtures

| fixture | kind | pins |
|---|---|---|
| `python-repo-symbol-not-imported-asks` | TP | real symbol never imported into the edited file |
| `python-module-function-called-without-qualifier-asks` | TP | `utils.slugify(...)` → bare `slugify(...)`; the import line still looks right |
| `python-symbol-imported-in-scope-clean` | TN | the edit's own import folds into scope |
| `python-hunk-scoped-out-of-scope-asks` | TP | control for both abstains — identical hunk, no marker |
| `python-star-import-abstains-clean` | TN | `from x import *` outside the hunk abstains the file |
| `python-dynamic-binding-abstains-clean` | TN | `globals().update(...)` outside the hunk abstains the file |
| `python-dynamic-binding-control-asks` | TP | same file minus the marker → asks |
| `go-file-scope-excluded-clean` | TN | v1 language gate |

Every true negative pins `enrolled_defs`, so its silence is a rule dropping a reachable candidate rather than an enrollment that quietly resolved nothing.

## Verification

- `go build ./...` clean, `gofmt -l` clean
- `TestHookCorpus` 21/21 pass; `./cmd/runecho-guard/...` and `./internal/guard/...` both ok
- mutation check above

## Remaining for #227

Contract fixtures (phase 5) — the last check, and the last new enrollment shape — then the bench statement of 6-of-6 coverage.


inth3shadows:
## Fixture set re-scored by mutation, not by argument

The first draft's 8 fixtures were chosen by reasoning about what they *should* cover. I scored them instead: break one behaviour of `internal/guard/filescope.go` / `cmd/runecho-guard/filescope.go` at a time, run the corpus, record which fixtures fail. A fixture that no defect can fail is a claim of coverage the corpus does not have — the exact failure #227 exists to fix.

**First draft — 4 of 8 caught nothing unique, and 3 real behaviours had no fixture:**

| behaviour broken | caught by |
|---|---|
| whole-file abstain arm | star-import, dynamic-binding |
| star-import gate | star-import |
| dynamic-binding gate | dynamic-binding |
| in-file scope check | imported-in-scope |
| edit-binds-names fold | imported-in-scope |
| imports bind in scope | imported-in-scope |
| flag gate | all 4 true positives (interchangeable) |
| **added-text abstain arm** | **nothing** |
| **no-whole-file-context bail** | **nothing** |
| **the firewall** | **nothing** |
| language gate (pure) | nothing |
| language gate (hook) | nothing |

`go-file-scope-excluded-clean` caught **zero** mutations — including the two language gates it existed to pin, since they are redundant with each other. Verified: removing *both* is still caught by nothing here.

**After the cuts and additions — 8 fixtures, all load-bearing, 10 of 12 caught:**

| fixture | sole catcher of |
|---|---|
| `python-symbol-imported-in-scope-clean` | scope check, edit-binds-names, imports-bind |
| `python-star-import-abstains-clean` | star-import gate (with the case below), whole-file arm |
| `python-dynamic-binding-abstains-clean` | dynamic-binding gate, whole-file arm |
| `python-star-import-added-by-the-edit-abstains-clean` | **added-text arm** |
| `python-new-file-no-whole-file-context-clean` | **no-context bail** |
| `python-firewall-leaves-invented-symbols-alone` | **the firewall** |
| `python-hunk-scoped-out-of-scope-asks` | flag gate (Edit shape) |
| `python-repo-symbol-not-imported-asks` | flag gate (Write shape) |

Cut: `python-module-function-called-without-qualifier-asks`, `python-dynamic-binding-control-asks`, `go-file-scope-excluded-clean`.

**Two harness fields, each forced by a case that was otherwise inexpressible.** `ask_without_flag` — a suppression rule is invisible to ask/no-ask, because removing the firewall only *doubles* an existing report; it asserts the always-on ask exists with the gate off and that the gate-on reason is byte-identical. `no_pre_edit_file` — an existing empty file still yields one blank line and the check runs; only a missing file engages the bail. The first version of that fixture used `"old": ""` and passed for the wrong reason.

**Known gap, now stated rather than implied:** the Python-only restriction is unpinned at hook level and covered only by the `internal/guard` unit suite.

Issue #209, opened 2026-07-23
contract: a repeat out-of-scope ask has no suppression (#12 D2 follow-up)


inth3shadows:
Found by adversarial review of #206.

A session declares `cmd/runecho-guard/**` and then legitimately has to touch `internal/guard/lang.go` twenty times — a refactor the contract did not anticipate, which is precisely the case `internal/contract/contract.go` calls "frequently legitimate". Each edit produces an identical ask naming the same path and the same contract.

The fact checks have two escape valves for repeat noise (`.runechoguardignore` and learned-allow). This dimension deliberately has neither, and its only documented remedy is to hand-edit the contract mid-session — which may itself be out of scope.

By the design's own standard ("noise of exactly the kind that trains a person to ignore the tool") this is the highest-volume ask source in the set with the weakest mitigation.

**Deliberately not fixed in #206.** This is the risk the plan names as D2's central one, and inventing a suppression mechanism before the dogfood window would be guessing at a number we are about to measure. Per `~/.claude/plans/runecho-12-contracts.md`, a high approve-anyway rate here is a *design* signal — possibly a signal to close the track — not a bug to patch.

Decide after the dogfood window, with `runecho-ir fpreport` data in hand. Options if it does need addressing:
- per-(session, path) once-only ask
- an ask that offers to widen the contract inline
- accept it, on the grounds that the friction is the feature

inth3shadows:
**Not fixing this yet — agreeing with the issue.** Building suppression before
the dogfood window is exactly the guess this issue says not to make, and a high
approve-anyway rate here is a signal about the track, not a bug to patch.

**But the gate could not have been read.** `fpreport` buckets on the *exact*
logged reason string, and reasons are compound — `askReason` joins co-firing
checks with `+`, `contractReason` prefixes `contract+`. So the contract check's
rate was spread across `contract`, `contract+violations`, `contract+dangling`…
with no per-check number available except hand arithmetic across buckets.

`contractReason`'s own doc note already says the arithmetic "has to split on `+`
and count terms". Nothing did — and this issue's gate depends on it.

#218 adds `FPStats.ByCheck`: the same tally with compound reasons split per
term, in the text report and the JSON payload. `ByReason` is kept (it shows
which checks *co-fire*, a real noise signal); the two totals deliberately
disagree, so `ByCheck` is documented as never summable to a window total.

**This issue is now measurable.** After a dogfood window with
[...]

Issue #12, opened 2026-06-06
Phase 4 Track D: contracts (V8 migration reserved)


inth3shadows:
**Order: 9 — after Phase 3**

D1: contracts table + YAML format (scope globs) — **consumes the reserved V8 migration**. D2: guard scope dimension (edit outside contract scope → ask). D3: contract registration hook. D4 (optional): GitHub Action, PR-description-derived contract vs PR diff — overlaps truth-trail's stale-claims output.

Source: runecho-phased-master-plan-2026-06.md (Phase 4).

inth3shadows:
Prioritized for next session (2026-07-18) alongside #170. NOTE: this stays `gated` — its "Order: 9 — after Phase 3" dependency isn't fully cleared (Phase 3 E5 duplicate-symbol is still open). Next session should either land the E5 dependency first or make an explicit call to ungate contracts before starting the V8 migration.

inth3shadows:
**Correction: the V8 reservation is stale — V8 is already consumed.**

`migrations` in `internal/snapshot/db.go` currently reads:

```
migrateV8, // 7 → 8: symbols.sig_hash — per-symbol body hash for modified-symbol diff
```

V8 was taken by `symbols.sig_hash` in e9e0a64 (the AST-accurate Python symbols + modified-symbol diff work), not by contracts. So D1 does **not** consume a reserved V8 — it would consume **V9** (or whatever is next when it lands).

Nothing about the design changes; this only corrects the premise so a future session doesn't go looking for a reservation that isn't there, or assume the migration slot is already claimed. Also unblocks #13, whose stated dependency was 'after contracts takes V8'.

inth3shadows:
**The Phase 3 gate appears to have cleared — this may be ready to start.**

This issue is labelled `gated` on "Order: 9 — after Phase 3". Phase 3 is E1 (dangling-refs) + E5 (duplicate-symbol) + E6 (auto-fresh IR). As of 2026-07-23 all three are shipped and dogfooding:

- `duplicateEnabled()` is live in `cmd/runecho-guard/main.go` (E5)
- `RUNECHO_GUARD_DANGLING`, `RUNECHO_GUARD_DUPLICATE` and `RUNECHO_GUARD_FILESCOPE` are all enabled in the dogfood environment
- E6 cleared its dogfood gate back in 2026-06-26

[...]

PR #255, opened 2026-07-29
fix(fpreport): count symbol-less asks instead of dropping them (#254)

### LINKED ISSUE #254: bench: fpreport is blind to contract-only asks — 16 of 19 are silently dropped
Found while dogfooding #12 under a live contract (PR #253).

## The gap

`FPReport` drops any hook ask with an empty `symbols[]` (`internal/guardstats/fpreport.go`):

```go
case "ask":
    if d.Mode != "hook" { continue }
    if len(d.Symbols) == 0 { continue } // no join signal
    asks = append(asks, d)
```

The skip is correct on its own terms — `symbolKey` is the join to outcomes, and a symbol-less record would collide with every other symbol-less record. But a **contract** ask carries no symbols. It is a statement about a *path*, not about identifiers.

Measured on the live log:

```
hook asks mentioning contract : 19
  ...with a symbols[] field   :  3   <- only these reach FPReport
  ...contract-ONLY (no syms)  : 16   <- silently dropped
  by reason: {'contract': 16, 'contract+violations': 3}

ALL hook asks: 649, dropped for empty symbols: 17
  dropped by reason: {'violations': 1, 'contract': 16}
```

The 3 that survive are `contract+violations`, and they survive only because the *violations* half supplied symbols. A pure contract ask is invisible.

## Why this is priority-1

The whole #12 Track D plan turns on one number. Per `~/.claude/plans/runecho-12-contracts.md` and #209, the decision to keep, suppress, or **close** the contracts track is supposed to be made from a dogfood approve-anyway rate read out of `runecho-ir fpreport`.

**That instrument cannot see the check.** `fpreport --json` today reports `{"check": "contract", "asks": 3, ...}` against 19 real contract asks — and those 3 are the ones that are not contract-only.

This is the same failure class as #227: a published measurement that silently covers a fraction of what it claims. There, five of six checks had no replayable fixtures; here, the FP report drops 84% of one check's asks and reports a number anyway.

Consequences:
- Any contracts ask rate read today is computed from a **non-random 16%** subset — specifically, the subset that co-fired with a symbol check, which is not representative of scope-only drift.
- #209's suppression design is explicitly waiting on this number.
- The `--max-rate` gate cannot ever gate the contract dimension.

## Scope

The join is the hard part, not the skip. Options, roughly in order of preference:

1. **Give contract asks a join key that isn't symbols.** The natural one is the file path plus the contract name — a contract ask is already keyed by path. That likely means widening the outcome record to stamp the same key, so the pairing stays exact rather than a time-window guess (which the plan explicitly warned against).
2. **Report contract asks as a separate bucket with volume only**, no rate, and say so — honest, and strictly better than the current silent 16% sample, but it does not produce the approve-anyway number #209 needs.
3. Leave it, and delete the claim that `fpreport` measures the contract dimension.

Whichever is chosen, `fpreport` should not print a `contract` bucket computed from a partial sample without saying it is partial.

## Do not

Do not read a contracts dogfood rate as evidence for or against closing #12 until this is resolved. That was the guidance I gave off #252, and #252 turned out to be the smaller of the two measurement problems.

Related: #12, #209, #252 (PR #253), #227.

inth3shadows:
Closes #254.

## The defect

`FPReport`'s join is symbol-exact, so an ask carrying no symbols had no key and
was dropped outright. The edit-scope contract check (#12 D2) fires on a **path**,
not on identifiers — so nearly every contract ask landed in that drop.

`fpreport` printed `{"check":"contract","asks":3,"approved":0,"rate":0}` — "the
contract check is never approved anyway" — while **13 real contract ask events
were invisible**. The 3 survivors are exactly the asks that co-fired with a
symbol-bearing check, which is the opposite of a random sample, presented with
nothing saying it was a sample at all. Same failure class as #227: a published
measurement silently covering a fraction of its own claim.

This mattered because the keep/close call on #12 Track D and #209's suppression
design both turn on that number.

## The fix

Symbol-less asks are now **counted but still never joined**.

- `FPBucket` gains `Unrated`, `Total()` and `Coverage()`.
- Window, reason, check, language and version buckets all carry unrated asks, as
  does the `LoudestRepos` volume ranking — answering a volume question while
  discarding real asks is the same defect at a different address.
- `FormatFP` marks a row `!` below 50% coverage, and prints **no percentage at
  all** for a bucket where nothing is rateable (rather than a `0%` that reads as
  "never approved"). A window of only unrated asks no longer reports itself empty.
- `PayloadFP` emits `unrated` / `total` / `coverage` beside `rate` on every bucket.
- `sortedFPKeys` orders by total, so an all-unrated check isn't buried.
- The outcome side keeps its drop. That is what makes counting the ask side safe:
  no symbol-less outcome is ever indexed for one to pair with.

`Window.Asks` keeps its old meaning — rateable asks — so the `--max-rate` gate's
denominator is untouched, pinned by a test.

## Rejected: a `(file, contract)` join key

Piloted on the reference log **before any code was written**, as the plan required.

A contract ask has no symbols, so that key collapses to one axis, and the outcome
recorder already pairs on file alone (`declog.go`'s `recentAsk`) with posture
always `ask`, never `deny` — "edited again" and "approved" become the same event.

The pilot bore it out. Aggregate 15.4%, which looks like discrimination until you
split it by contract:

| contract | ask events | would pair | |
|---|---|---|---|
| `guardstats-dedup` | 2 | 2 | the only real dogfood task — **100%** |
| `contracts-d2` | 10 | 0 | D2 dev session, synthetic files (`x.md`, `brand_new.go`) |
| `d3a-acceptance` | 1 | 0 | synthetic acceptance probe (`probe_d3a.go`) |

That worktree logged 13 asks and **0** outcomes on a day the recorder wrote 23
elsewhere — at the pooled 79% rate, 0-for-13 is not a denial signal; those were
manual guard invocations with no PreToolUse prompt, so no PostToolUse was ever
possible. Strip them and the join is 100% by construction on n=2.

It would also have had to forward `decisionRecord.Contract` onto the outcome
record, which `declog.go` deliberately does not do (*a scope that trains itself
wider is not a scope*). Neither price buys a number worth reading.

## Measured

Reference log, `--days=90`, 20,936 records:

- Overall **unchanged**: 556 asks / 439 approved = 79.0%.
- `contract` now reads `3 ask  0 approved  0%  ! +13 unrated (rate covers 19%)`.
- 14 unrated asks surfaced in total (13 contract, 1 violations).

## Verification

Full suite green, `go vet` clean, `gofmt` clean under the go.mod-pinned 1.25
toolchain. 8 new tests, each mutation-scored:

| mutation | killed by |
|---|---|
| restore the `len(Symbols)==0` ask drop | 8 tests |
| count unrated into `Asks` instead of `Unrated` | 7 tests, incl. the gate-denominator test |
| lift **both** symbol-less guards without a real key | collision fixture — `approved = 1` |
| minute-resolution collapse key | contract-ask triplication test |
| drop the `!` low-coverage marker | coverage-disclosure test |
| empty-window check on `Asks` not `Total` | only-unrated-window test |
| payload drops `coverage` | payload test |
| order buckets by rateable asks | ordering test |

## Note on #12

The instrument is fixed; the standing instruction stands for a different reason.
There is still only **one** real dogfood contract task on record. Don't read a
contract approve-anyway rate off this report yet — the row now says why itself.


inth3shadows:
## Review round 1 — all 7 findings fixed (`679d9d5`)

One was a **real regression this PR introduced**, and it was in the one thing the PR body claimed was untouched.

**#1 (MEDIUM) — gate eligibility.** Unrated asks create `ByVersion` buckets, and `MixedVersions()` counted map *keys*. So one symbol-less ask from a different build — or a single pre-symbol-stamping record, which lands under `UnknownVersion` and exists on every long window of the real log — flipped it true and, through `gateEligible`, **silently skipped the `--max-rate` gate** on a window whose rate was single-version. Verified: 25 rateable asks on `v0.17.18` + 1 unrated ask on `v0.17.10` took `mixed` false → true. It also printed "the rate above pools 2 different builds" about a rate drawn from one.

`MixedVersions()` now counts versions **with a rateable ask** — a bucket that contributed to no rate cannot make a rate an average. The PR pinned `Window.Asks` as the untouched denominator; nothing pinned eligibility. That test exists now.

**#2 (MEDIUM)** — the headline printed `0% approval rate` for an all-unrated window, the exact misreading `fpRow` was added to prevent, one level up and *above* the caveat. Now `0 rateable ask(s) → no approval rate`. Banner wording fixed too ("N further asks" with nothing to be further than).

**#3 (MEDIUM)** — `PayloadFP` emitted `"rate": 0` where `FormatFP` refused to print one, keeping the defect on the machine path `PayloadFP`'s own doc names as at risk. `null` now: a consumer that plots null plots a gap; one that plots 0 plots a claim.

**#4 (MEDIUM)** — three different figures for one log across comment / doc / USAGE / tests (22 & 14% & `rate 1.0` vs 16 & 19% vs 19 & 15.8%), on the number the #12 keep/close call is meant to be made from. All prose now carries the measured value, dated, with a re-measure warning.

**#5 (LOW)** — `%5.0f` rendered 0.9978 as `+1 unrated (rate covers 100%)`. Truncates now, with `>99%` / `<1%` forms.

**#6 (LOW)** — the `ask` column meant `Total` in the no-rate branch and `Asks` elsewhere; and the 0%-rated row was the one row missing the `!` the banner tells you to look for. Column is always the rateable count; marker moved outside the branch.

**#7 (LOW)** — `loudest_repos[].asks` blended rated and unrated, giving one JSON name two denominators in one document. `RepoCount` splits `Asks`/`Unrated`, ranks on `Total()`, and a test pins that `sum(loudest_repos[].asks)` reconciles with `overall.asks`.

### Verification

6 new regression tests, each mutation-scored — reverting a fix kills that fix's test and no other:

| reverted fix | killed |
|---|---|
| `MixedVersions` counts map keys | `MixedVersionsIgnoresUnratedOnlyVersions` |
| headline prints rate unconditionally | `HeadlineSuppressesRateWhenNothingIsRateable` |
| payload `rate` always numeric | `FullyUnratedBucketHasNullRate` |
| coverage rounds with `%.0f` | `CoverageNeverRoundsUpToFull` |
| `RepoCount.Asks` blends unrated | `LoudestReposSplitsRatedAndUnrated` + `ContractOnlyAsksAreCountedNotDropped` |

Full suite green, `go vet` clean, `gofmt` clean under the go.mod-pinned 1.25 toolchain. Reference log unchanged where it should be: 561 asks / 441 approved = 79.0%; contract at 3 rated / 13 unrated / 18% coverage.

Issue #243, opened 2026-07-24
guard: call-shape agreement — verify a resolved symbol is *used* correctly (arity/kwargs/return), not just that it exists


inth3shadows:
## Problem

The guard verifies a symbol **exists** (`does foo exist`). It does not verify the agent **uses it correctly** (`does this call agree with how foo is declared`). This is the highest-damage hallucination class that survives every existing check: the symbol resolves, the guard passes, and the agent still ships `foo(a, b)` when the declaration is `foo(ctx, a, b) (T, error)`. It *looks* grounded, so nothing downstream flags it until compile/runtime.

Failure taxonomy of an AI coding agent, and who catches each:

| Failure | Caught by |
|---|---|
| Invents a symbol | RunEcho (existence) ✓ |
| Token bloat / stale context | terse ✓ |
| Edits blind to blast radius | CodeGraph ✓ |
| **Uses a *real* symbol wrong — arity, keyword args, return handling** | **nobody — this issue** |
| Wrong logic / computes wrong thing | tests/execution (not static) |

Existence is necessary but shallow; **agreement** is where the value is, and it is still deterministic, in-loop, and model-free — RunEcho's whole lane.

## Scope — what a deterministic guard CAN check without a full type system

Cheap, high-signal, no type inference required:
- **Arity** — positional-arg count vs declared params (min/max accounting for defaults, varargs/`*args`, rest params).
- **Keyword/named args** — a kwarg or named field the declaration doesn't accept (`foo(timeuot=...)`, unknown struct field).
- **Return-arity** — Go multi-return mismatch (`x := foo()` when `foo` returns `(T, error)`); Python tuple-unpack width.
- **Call vs non-call** — invoking a non-callable, or referencing a function value where a call is expected.

Explicitly OUT of scope (needs a type system → that's `lsp-go`, not the deterministic guard):
- Argument *type* correctness (`int` vs `string`).
- Overload/generic resolution.
- Interface-satisfaction.

The line: RunEcho owns **shape** agreement (counts, names, callability) deterministically; the LSP bridge owns **type** agreement. Don't drag a type-checker into the hook.

## Relationship to existing work

- NOT #12 (contracts) — that is *file-scope* globs (which files an edit may touch), a different dimension.
- Complements `lsp-go` (decision #291): LSP gives type-aware depth for one language, heavy/stateful; this stays language-agnostic and cheap for the shape layer runecho already parses.
[...]

PR #258, opened 2026-07-29
guard: extract Python call shape from a diff (#243 step 1)

inth3shadows:
#243 step 1. **Stacked on #257** — base is `fix-fstring-nested-literal`, not `master`, because the differential test only reaches 100% agreement with that fix in place. Merge #257 first; this PR's base then retargets to `master`.

No check is wired, no schema change. This is the call-side half only.

## What it does

`ExtractCallShapes` returns, for every unqualified call site whose argument list is **fully visible in the added lines**, the positional count, the keyword names, and flags for the constructs that make a shape unknowable (`*args`, `**kwargs`, a top-level `lambda`, an unclassifiable segment).

**Abstention is the design.** A call whose argument list does not balance inside the added lines is skipped, not guessed at — arguments continuing into unchanged context above or below the hunk are unknowable from a diff, as are those interrupted by a hunk gap or past the byte/line ceilings. Consumers see only shapes that are certain, which is what lets a future mismatch be reported as fact rather than suspicion.

## Precision, measured against CPython's `ast`

The step-2 spike used a perfect AST on *both* sides, so it proved reach and comparison logic but explicitly could **not** measure this extractor's own false-positive rate. `TestCallShapeDifferential` does: it compares emitted shapes against CPython, and only emitted sites are compared, since an abstention is safe by design.

| corpus | oracle sites | usable | recall | **agreement** |
|---|---|---|---|---|
| NFLsignal (211 files) | 7898 | 4723 | 59.8% | **100.00%** |
| boundary-probe (56 files) | 1749 | 1275 | 72.9% | **100.00%** |
| `testdata/callshape` (CI) | 36 | 25 | 69.4% | **100.00%** |

Reproduce with `RUNECHO_CALLSHAPE_CORPUS=<repo> go test ./internal/guard/ -run TestCallShapeDifferential -v`. The test skips when `python3` is absent, so CI without Python still passes on the checked-in corpus.

## Two defects it found that the unit suite had passed

1. **The f-string masking bug** — split out as #257 (the parent commit), because it is a pre-existing false positive in the *shipped existence check*, not new here.
2. **A masked string-literal argument is byte-identical to the nothing after a trailing comma**, so `_fmt(gap, '+.4f')` read as ONE argument. Structural decisions now run on masked text while argument *presence* is judged on the index-aligned source. Where the two cannot settle it — a comment interleaved with the arguments, indistinguishable from a value on the next line once the list is joined — the shape sets `Unreliable` rather than undercounting. That cost 5 of 9647 sites.

Both are the same failure mode: masking is length-preserving and therefore lossy about *why* a span is blank. Where that ambiguity is load-bearing, the extractor now says so instead of picking.

## Review focus

- `keywordName` is the FP-critical function: `==`, `!=`, `<=`, `>=`, the walrus `:=`, a nested lambda default, and any `=` inside brackets must all read as positional. Each has its own assertion.
- `collectArgText` returns masked and original text that **must stay index-aligned**; it abstains if the invariant breaks rather than mis-slicing.
- Ceilings (`callShapeMaxLookahead`, `callShapeMaxArgBytes`, `callShapeMaxSites`) are DoS-relevant per the #212 red-team pass — the guard scans attacker-influenced repo text inside an edit loop.

## Deliberately not here

Argument *type* correctness, overload/generic resolution, interface satisfaction — permanently out of scope, `lsp-go` owns those (KB decision #291). Also no `migrateV10`: the declaration side resolves on demand in the next slice, so the FP and latency questions get answered before any schema is committed. Plan and the full step-2 spike record: `~/.claude/plans/runecho-243-call-shape-agreement.md`.

## Known reach cap

10.6% of kwarg-bearing call sites in the larger corpus must abstain because the index is name-keyed and cannot say which same-named declaration is in scope. Indexing the shape would not lift that — it needs import/scope resolution. Recorded, out of scope for this slice.

## Verification

- `go test ./...` green
- `gofmt -l ./cmd ./internal` clean (Go 1.24 toolchain, to match CI)
- `go vet ./...` clean
- `bench/hookmutate/mutate.py --strict` exits 0 — 36 mutations / 28 fixtures, the 6 survivors all pre-existing with documented known-gap annotations


inth3shadows:
Superseded by #259 — same branch and commits, rebased onto `master`. GitHub auto-closed this PR when its base branch (`fix-fstring-nested-literal`, merged as #257) was deleted, and refuses to reopen or retarget a closed PR.

Issue #145, opened 2026-07-15
guard: ExtractRefs seeds string state from the hunk, so lines added inside a pre-existing docstring are scanned as code


inth3shadows:
## Summary

`ExtractRefs` (`internal/guard/extract.go`) masks string/docstring content via `stripLiteralsStateful`, threading multi-line state through `open`. But `open` is seeded to `""` from the *added lines alone* — the guard never sees the unchanged lines above the hunk. When a diff adds lines **inside a pre-existing docstring** (the opening `"""` is unchanged, so it is not in the hunk), the state machine starts in the "outside string" state and scans the added prose as code. `reCallIdent` then reads any `word (` in that prose as a call, and the guard blocks the commit on unresolved English words.

This is the diff-scoped complement to the masking that already works correctly for a full-file scan.

## Evidence (current `master`, acd2fb0)

`extract.go:497-503`:
```go
seen := make(map[string]struct{})
// open tracks an unterminated multi-line string delimiter carried across
// lines (Python triple-quote `"""`/`'''`, JS/Go backtick). It resets on a
// non-consecutive line, since a diff hunk's added lines may not be contiguous
// and string continuity can't be assumed across a gap.
open := ""
prevNo := 0
```

`open` starts empty and is only ever advanced by lines *within* the hunk, so a hunk beginning inside a docstring is indistinguishable from one beginning in code. The masking at L529 is then a no-op for those lines:

```go
scan, newOpen := stripLiteralsStateful(lang, text, open)
```

`reCallIdent` (L359) matches any identifier followed by `(`:
```go
var reCallIdent = regexp.MustCompile(`([A-Za-z_$][\w$]*)\s*\(`)
```

## Live repro

Real commit in `inth3shadows/secret-broker` (2026-07-15), adding prose to the **module docstring** of `src/secret_broker/ops/_runner.py`. Line 1 of that file is `"""Shared HTTP-proxy plumbing: ...` — unchanged, therefore absent from the hunk. Guard output:

```
[runecho-guard] 3 unresolved symbol(s):
  src/secret_broker/ops/_runner.py:28: the  (did you mean "re"?)
  src/secret_broker/ops/_runner.py:32: metadata
  src/secret_broker/ops/_runner.py:288: headers
```

The three source lines are ordinary prose, each with a word before a parenthetical:

| Line | Docstring text | Parsed as |
|---|---|---|
| 28 | ``The redactor sees the (projected) response body string`` | `the(...)` |
[...]

PR #259, opened 2026-07-29
guard: extract Python call shape from a diff (#243 step 1)

inth3shadows:
#243 step 1. **Supersedes #258**, which GitHub auto-closed and refuses to reopen when its base branch (`fix-fstring-nested-literal`, merged as #257) was deleted. Same branch, same commits, rebased onto `master`.

No check is wired, no schema change. This is the call-side half only.

## What it does

`ExtractCallShapes` returns, for every unqualified call site whose argument list is **fully visible in the added lines**, the positional count, the keyword names, and flags for the constructs that make a shape unknowable (`*args`, `**kwargs`, a top-level `lambda`).

**Abstention is the design.** A call whose list does not balance inside the added lines is skipped, not guessed at — arguments continuing into unchanged context are unknowable from a diff, as are those interrupted by a hunk gap or past the ceilings. Consumers see only shapes that are certain, which is what lets a future mismatch be reported as fact.

## Precision, measured against CPython's `ast`

The step-2 spike used a perfect AST on *both* sides, so it could not measure this extractor's own FP rate. `TestCallShapeDifferential` does.

| corpus | oracle sites | usable | recall | **agreement** | unreliable |
|---|---|---|---|---|---|
| NFLsignal (211 files) | 7898 | 4728 | 59.9% | **100.00%** | 0 |
| boundary-probe (56 files) | 1749 | 1275 | 72.9% | **100.00%** | 0 |
| `testdata/callshape` (CI) | 36 | 25 | 69.4% | **100.00%** | 0 |

Reproduce with `RUNECHO_CALLSHAPE_CORPUS=<repo> go test ./internal/guard/ -run TestCallShapeDifferential -v`. Skips when `python3` is absent.

## Review history — this branch was reviewed and five findings fixed

`/code-review` found real defects, including one I had shipped in the first commit. All are addressed here or in #257:

1. **HIGH (fixed in #257)** — `maskNestedLiteral` blanked a nested *f-string*, whose interpolation is code, dropping real calls. `f"outer {f'{Build(x)}'}"` returned `[]` where master returned `[Build]` — a false negative, the worse direction. My first fix attempt passed two cases and corrupted a third.
2. **Total-scan budget** — the per-attempt ceilings bound one list and `callShapeMaxSites` bounds successes, but an *abandoned* attempt costs a full lookahead and yields no site. Unbalanced input measured 19 s. Now budgeted, and the budget value is set from measurement: 4 MiB cost 45 ms against a ~12 ms envelope, so it is 256 KiB at 5.1 ms.
3. **`openSeed`** — without it, added lines inside a pre-existing docstring emitted prose (`COUNT`, `helper`) as *certain* shapes: the #145 class the existence check already fixed. Parameter added before any consumer is wired.
4. **`Unreliable` on ordinary commented Python** — the stated rationale was refuted by this branch's own sibling test. A `#`-leading segment is provably a pure comment. Fixing it recovered the 5 previously-abstained sites.
5. **The string-literal branch was unpinned and its rationale wrong** — `stripLiteralsStateful` keeps quote bytes, so in plain context the branch is unreachable; it is reachable only inside an f-string interpolation, which is now the case the test uses.

Plus: a flaky wall-clock assertion in my own budget test (failed CI at 6.05 s under `-race`) replaced with a behavioral one, the contract file's quoted description and missing self-scope, an orphaned godoc comment, and a duplicated doc block.

## Review focus

- `keywordName` is the FP-critical function: `==`, `!=`, `<=`, `>=`, walrus `:=`, a nested lambda default, and any `=` inside brackets must all read as positional. Each has its own assertion.
- `collectArgText` returns masked and original text that **must stay index-aligned**; it abstains if the invariant breaks rather than mis-slicing.
- `Unreliable` is now a defensive default with **no known trigger** and deliberately no fixture — flagging that explicitly rather than claiming coverage.

## Deliberately not here

Argument *type* correctness, overload/generic resolution, interface satisfaction — permanently out, `lsp-go` owns those (KB decision #291). No `migrateV10`: the declaration side resolves on demand in the next slice, so FP and latency get answered before any schema is committed. Plan and full spike record: `~/.claude/plans/runecho-243-call-shape-agreement.md`.

## Known reach cap

10.6% of kwarg-bearing sites in the larger corpus must abstain because the index is name-keyed and cannot say which same-named declaration is in scope. Indexing the shape would not lift that — it needs import/scope resolution. Out of scope for this slice.

## Verification

- `go test -race ./...` green
- `gofmt -l ./cmd ./internal` clean (Go 1.24 toolchain, to match CI)
- `go vet ./...` clean
- `bench/hookmutate/mutate.py --strict` exits 0 — 36 mutations / 28 fixtures, 6 survivors all pre-existing with documented known-gap annotations


inth3shadows:
## Review 2 — six confirmed defects fixed. **Not merged.**

The review pointed this PR's own differential harness at the **CPython 3.12 stdlib** (23856 call sites) and got **99.88%**, not the 100% my three corpora reported. Twelve wrong shapes, none flagged — so a consumer could not have abstained. The earlier 100% figures were corpus-selection artifacts: `testdata/callshape` contained none of the offending constructs, so CI could not falsify the claim. That is [the corpus-shares-your-assumptions lesson](https://github.com/inth3shadows/runecho/issues/256) recurring, now on my own harness.

All six root causes are fixed in `d30180a`, and **all four constructs are now in `testdata/callshape`** so each finding is a red test without the external corpus.

| # | Sev | Defect | Fix |
|---|---|---|---|
| 1 | HIGH | Unparenthesized genexp puts a comma at depth zero (`_sum(w / x for w, x in zip(a, b))` → `Pos=2`, truth 1). In `statistics.py`, `dataclasses.py`. | depth-zero `for` abstains, like `lambda` |
| 2 | HIGH | Method chain across a backslash continuation emitted an **unqualified** `replace`/`decode`. 6 of the 12. | joiner carried as run state |
| 5 | MED | `match`/`case` soft keywords read as calls; `case Point(x=0)` is MatchClass | pattern lines skipped |
| 3 | MED | `collectArgText` counted depth, not bracket TYPE — `foo(a, b]` returned a confident shape | bracket stack, abstain on mismatch |
| 4 | MED | PEP 701 same-quote nested f-string undercounted `Pos` silently | any surviving f-string prefix abstains |
| 7 | LOW | Lone backslash before `)` counted as an argument | treated as punctuation |

Plus **8** (the budget's doc comment reasoned from the hook envelope while `scanned` is a local — scope now stated, and a per-file consumer is told to thread its own) and **9** (the oracle captures stderr and SKIPs; `ast.parse` can raise `ValueError` on a NUL byte or `RecursionError`, and an oracle that cannot run is no evidence either way).

**A bug worth calling out on its own:** `hasTopLevelLambda`'s fast path tested `s[i] != 'l'`. Correct for the one keyword it was written for, and it silently made every *other* keyword unmatchable — so finding 1's fix appeared to do nothing until that line changed.

### Agreement now: 100.00% on all four corpora, 16033 usable shapes

| corpus | oracle sites | usable | recall | agreement | unreliable |
|---|---|---|---|---|---|
| `testdata/callshape` | 45 | 31 | 68.9% | **100.00%** | 2 |
| NFLsignal (211 files) | 7898 | 4611 | 58.4% | **100.00%** | 117 |
| boundary-probe (56) | 1749 | 1197 | 68.4% | **100.00%** | 78 |
| **CPython 3.12 stdlib** | 23856 | 10194 | 42.7% | **100.00%** | 134 |

**Recall fell and that is the intended trade.** NFLsignal went 4728 → 4611 usable with 117 abstentions, almost all from the conservative f-string rule (finding 4). A shape withheld costs a check that never fires; a wrong shape costs the guarantee.

`go test -race ./...` green, `gofmt -l ./cmd ./internal` clean (Go 1.24), `go vet ./...` clean.

PR #260, opened 2026-07-30
guard: report keyword arguments a declaration does not accept (#243 slice 1)

inth3shadows:
#243 slice 1: the guard now checks that a resolved symbol is **used** as it is declared, not merely that it exists. Python, keyword arguments only. Behind `RUNECHO_GUARD_CALLSHAPE=1`, ask-posture, default OFF.

Consumes `ExtractCallShapes` from #259 (the call-side half) and compares it against the same file's own module-level declarations. **No store I/O, no schema change, no new dependency.**

```
def fetch(url, timeout=10): ...
fetch("u", timeuot=5)
  -> [runecho-guard] 1 keyword argument(s) the declaration does not accept —
     the symbol resolves but the call does not match it:
       snippet line 5: fetch(timeuot=…)  (did you mean "timeout"?) —
       fetch is declared at line 1 and accepts: url, timeout
```

## Three design reversals, each on a measurement

**1. Same-file resolution only. No DB query.** A cross-file match would be *name-keyed*: nothing in the index can say whether this file's `foo` is the repo's `foo` or one imported from a third-party package of the same name, and answering that is import resolution — out of scope per #243, and the same ceiling the step-2 spike recorded.

Measured cost (CPython-`ast` oracle, venv-excluded):

| corpus | kwarg-bearing bare calls | same-file reach | same+cross | same-file share of checkable |
|---|---|---|---|---|
| NFLsignal (211 files) | 1431 | **21.2%** | 39.8% | 53% |
| boundary-probe (56) | 389 | **3.3%** | 26.7% | 13% |
| harness (16) | 20 | **15.0%** | 20.0% | 75% |

Roughly half the reachable population is given up. Accepted: the half that remains is the half where the answer is a fact.

**2. Tree-sitter cannot run at hook time. At all.** The declaration side was built twice on a real AST and deleted. On this project's pure-Go runtime, per candidate call on a 3037-line module:

| implementation | cost |
|---|---|
| whole-file parse | **180 ms** |
| 4-line signature snippet | **8 ms** — fixed per-`Parse` overhead, not size-dependent |
| masked scanner (shipped) | **0.97 µs** |

The hook's budget for everything it does is ~12 ms. No amount of snippet-narrowing gets under it. What ships reads the parameter list the way the call side already reads an argument list, sharing `splitTopLevel`/`isIdent`, so both halves use one set of rules. `internal/parser/pydecl.go` is deleted.

That is only defensible with an oracle: `TestCallShapeDeclDifferential` compares the accepted-keyword set, the `**kwargs` flag and the decorated flag against CPython. **4548 module-level declarations across 3 corpora, 0 disagreements** (NFLsignal 960 · CPython 3.12 stdlib 3127 · boundary-probe 461; 3 explicit abstentions, 2 not located).

**3. The file-scope check's abstain list is wrong for this check.** Reusing `pyDynamicBinding` (`setattr(`, `vars(`, `importlib`, `__import__`) abstained **218 of 283** checkable stdlib sites — one `getattr` silenced a 2000-line file. Those matter to "is this name bound anywhere"; they cannot matter to "does this def's parameter list accept this keyword". Narrowed to `globals(`/`locals(`/`exec(`/`eval(`, tested on masked text. Stdlib kill rate 38% → 71%, no new false positive.

## Abstain ladder as shipped

`HasLambda`/`Unreliable` · no keyword args · star-import or `globals(`/`locals(`/`exec(`/`eval(` in the file · callee bound by an import, an assignment, or any def's parameter · no column-zero `def` of that name · two column-zero defs disagreeing · declaration decorated · declaration takes `**kwargs` · signature unreadable · **this edit rewrites or deletes the declaration**.

That last one is not in the original plan and is the most valuable in the set: "add a parameter and pass it" is a routine agent move, and comparing against the pre-edit signature would false-positive on every one. The added text takes precedence when it carries a whole `def` line; a removed line belonging to the pre-edit parameter list abstains.

## Two narrowings, recorded rather than hidden

1. **Column-zero declarations only.** A module function nested under `if TYPE_CHECKING:` or `try:`/`except ImportError:` is indented, so its calls abstain. A column test is the only way to separate a module function from a method without a whole-file parse.
2. **Parameter shadowing is file-wide.** A parameter named `fetch` anywhere abstains every `fetch(...)` in the file; deciding whether the call sits inside that function needs scope analysis.

## The mutation gate changed the code, not just the fixtures

`bench/hookmutate/mutate.py --strict` exits 0 — no unclaimed survivor, no dead fixture. Getting there took three real changes:

- **An abstention that was simply wrong.** `call-kwstar-abstain` was unkillable because the rule was unjustified: `foo(bogus=1, **cfg)` is a TypeError exactly as `foo(bogus=1)` is — a splat *adds* keywords, it never excuses an explicitly named one. Removed; the oracle stopped skipping splat sites. **A splat does defeat an arity check, so this does not carry to slice 2.**
- **A redundant guard deleted rather than annotated.** `matchesPyDef(removedText, name)` survived even with a fixture written for it, because `removesSignatureLine` subsumes it — removing a declaration necessarily removes a line of its own signature.
- **Two fixtures deleted** per #227's rule (a fixture no defect can fail is a coverage claim the corpus does not have), three given mutations that make them earn their place.

Also learned: `edit-adds-param-and-uses-it-clean` does **not** pin added-text precedence, which its description claimed — the removed-text guard shadows it.

## Adversarial review found two real defects

Standing rule for guard-correctness changes; both reproduced end-to-end before being fixed.

**FALSE POSITIVE — a decorator outside the hunk was invisible.** An Edit whose `old_string` begins at the `def` line leaves the `@decorator` in the unchanged file, so `decoratedAbove` had nothing to walk, reported undecorated, and compared a call against a signature a wrapper may have replaced. Reproduced with `@app.route("/x")` plus a parameter-adding hunk.

**Why nothing caught it: every decorator test and the entire CPython-oracle harness use the Write posture, where the decorator is always visible.** No test could reach the Edit branch. A posture the harness cannot construct is a posture with no coverage, however many corpora it sweeps. Fixed by folding the pre-edit file's decoratedness in; an edit that *removes* a decorator now over-abstains for one edit.

**FALSE NEGATIVE — markers matched as bare substrings.** `def get_locals(a, b)` contains `"locals("`, so the file's dynamic flag went true and the check returned nil for the **whole file**. Any helper ending in locals/globals/exec/eval silently switched off its own module. Now requires a word boundary, and excludes a preceding `.` — `self.eval(...)` is a method call, which cannot rebind a module-level name.

Plus: the cheap prefilter tested `"def "` while both regexes accept `def[ \t]+`, so `def\tname(` was invisible to resolution and absent from the shadow set.

Checked and clean: unicode identifiers, walrus, PEP 604 unions, string defaults containing `#`/quotes/brackets, comments inside a parameter list, backslash continuations, bounds/nil in `signatureContent`/`pySignatureLines`/`addedIndex`, and the `main.go` wiring including `addedIsWholeFile` for MultiEdit and file-creating Writes.

## Verification

- `go test -race ./...` green · `go vet ./...` clean · `gofmt -l` clean under the go.mod-pinned Go 1.24 toolchain (local is newer and produces phantom CI diffs)
- `bench/hookmutate/mutate.py --strict` exits 0 — 17 call-shape fixtures, 13 call-shape mutations; the six remaining survivors are all pre-existing documented known gaps
- **False positives: 0** over 612 oracle-checkable call sites in four corpora (`TestCallShapeCheckNoFalsePositives`); one CONFIRMED true positive in the CI fixture corpus
- **Mutation kill: 100% / 99.0% / 100% / 70.9%** (CI / NFLsignal / boundary-probe / stdlib) — the anti-vacuity half, since 0 reports on clean code is equally consistent with a dead detector
- **Latency: 6.2 ms** worst case (3037-line file, candidate present), **4.3 µs** with no kwarg-bearing call in the hunk. Three reorderings were needed: candidates narrowed before any whole-file work, a `"def"` substring reject before the two regexes, and the shadow set built on first use
- Python's grammar ships in **both** channels already (`install.sh` `GRAMMAR_TAGS` and `.goreleaser.yaml`) — verified, not assumed. This slice adds no grammar dependency anyway.

Reproduce the corpus numbers with `RUNECHO_CALLSHAPE_CORPUS=<repo> go test ./internal/guard/ -run 'TestCallShapeCheck|TestCallShapeDeclDifferential' -v`. Skips when `python3` is absent.

## Not in this PR

Argument *type* correctness, overload/generic resolution, interface satisfaction — permanently out, `lsp-go` owns those (KB decision #291). Positional arity, return arity and call-vs-non-call are later slices. Default-on waits on a dogfood window and `runecho-ir fpreport`.

Plan and full decision record: `~/.claude/plans/runecho-243-call-shape-agreement.md`.


inth3shadows:
## `/code-review` round 2 — three reproduced false positives, fixed

All three were reproduced by running the code, not argued from reading it. Commits `f0d4…` onward.

**1. Only plain `=` assignments counted as rebinding.** `pyNonDefBindings` derived its shadow set from `PyDeclaredNames`, which sees only `pyAssignLHS`. Every other Python binding form left the callee looking unshadowed, so the check compared against a module-level def the call never reaches. Five reproductions on valid code:

```python
for handler in fns: handler(item, verbose=True)      # def handler(x, dry=False)
with make() as opener: opener(p, binary=True)
except Exception as fetch: fetch("u", verify=False)
[fetch("u", verify=False) for fetch in fns]
if (fetch := g()) is not None: fetch("u", verify=False)
```

Added `rebindings()`, lazy, off the already-masked lines. The `for` pattern is deliberately **not** the anchored `rePyForTarget` from `dropped_import.go` — a comprehension puts the `for` mid-line, which is exactly what an anchored pattern misses.

**2. An indented redefinition was invisible.** Only column-zero defs entered `defLines`, so a nested `def process(rows=None)` inside a function — or a conditionally-declared `if True: def f(b=2)` — was not seen beside a column-zero def of the same name, and the call was compared against the wrong signature. The documented narrowing claimed such a call abstains; that held only when there is *no* column-zero def, which is not the failing case.

A **method** still does not shadow, and that exclusion is load-bearing: an unqualified `fetch(...)` cannot reach `C.fetch`, so treating every indented def as a shadow would abstain on a method and a module function sharing a name — one of the most common shapes in Python. The test is the nearest enclosing header at strictly smaller indentation.

**3. The signature span was scanned on RAW text.** `pySignatureLines` counted parens on raw lines while every other signature reader uses masked text, so `sep="("` in a default made the span never balance → returned nil → `removesSignatureLine` false → the "this edit rewrites the declaration" abstain silently stopped working, reporting a correct parameter rename as a mismatch. Replaced by `signatureLinesAt`, balancing via the masked `signatureContent`.

**4. The oracle was blind to the class it was certifying.** It ignored rebinding, so finding 1's false positives would have scored as `CONFIRMED` true positives — the "0 false positives" headline structurally could not see them. It now skips any callee rebound by a store-context binding, a parameter, an import, or a non-column-zero def. Over-skipping only shrinks the population; under-skipping corrupted the claim. **Stdlib checkable sites 284 → 265: the number got smaller because it got honest.**

**5.** The declaration line's coordinate system is now labelled — a shape read from the added text is numbered within the *hunk*, and printing it as a file line sent the reader elsewhere. Prints `snippet line` in that case, matching the call site's existing label.

**6.** The benchmarks no longer read `/usr/lib/python3.12/...`. They `b.Skipf`'d on CI, on macOS, and after any Python upgrade, so every latency number the design rests on would have quietly become unmeasurable. The corpus is generated in-process.

### Re-measured

- **0 false positives** over 593 oracle-checkable sites, four corpora
- **4564 declarations vs CPython `ast`, 0 disagreements**
- mutation kill **100% / 99.0% / 100% / 74.5%**
- **5.6 ms** worst case, **3.4 µs** with no kwarg-bearing call
- `go test -race`, `go vet`, Go 1.24 `gofmt` clean; `mutate.py --strict` exits 0 — 20 call-shape fixtures, 16 call-shape mutations, no dead fixture, no unclaimed survivor

### Known gap, not fixed here

The check is wired **after** `main.go`'s store gate, so an unenrolled tree or a snapshot-less store skips a check that needs no store data at all. Contracts have an `askContractOnly` pre-gate for exactly this; call-shape has no equivalent. Default-off makes it low priority, but it is a real coverage gap and is recorded rather than implied away.

### Also worth stating plainly

The first review round's lesson repeated itself: **a posture the harness cannot construct is a posture with no coverage.** Round 1 found the decorator FP because every decorator test used the Write posture. Round 2 found the rebinding FPs because the oracle modelled no rebinding at all. Both times the corpus number was real and blind at the same time.

Issue #268, opened 2026-07-31
guard: three checks have no reason bucket — file-scope, qualified and deps-go are logged as "violations"


inth3shadows:
Found while reading the dogfood log to answer whether the gated checks are ready to un-gate. They are not answerable, and this is why.

`askReason` (`cmd/runecho-guard/dangling.go:229-249`) emits exactly five terms: `violations`, `dangling`, `dropped-import`, `duplicate-symbol`, `call-shape`. But three more checks append their findings straight into the `violations` slice:

| check | flag | folded in at |
|---|---|---|
| file-scope resolution | `RUNECHO_GUARD_FILESCOPE` | `main.go:735` (hook), `main.go:281` (pre-commit) |
| same-repo qualified | `RUNECHO_GUARD_QUALIFIED` | `main.go:715`, `main.go:252` |
| dependency qualified | `RUNECHO_GUARD_DEPS_GO` | `main.go:725`, `main.go:267` |

`guardstats` and `fpreport` bucket on the exact reason string, so all three are invisible as themselves and are counted as the additive hallucination check.

**Two consequences, and the second is the worse one.**

*Their own rate does not exist.* `RUNECHO_GUARD_FILESCOPE=1` has been on since ~2026-07-22 and the log holds 9,095 Python decisions, more than any other language. Reading the log naively says file-scope has fired zero times. That is not a precision result — there is simply no bucket for it to land in. The un-gating criterion for these checks ("an fpreport window shows the false-positive rate on real edits") cannot be evaluated, and no amount of further dogfooding will change that.

*They contaminate the check that IS default-on.* `fpreport -days 30` reports `violations` at 344 asks / 74% approve-anyway. That number is a blend of the additive check and up to three gated ones, in unknown proportion. It is the headline number for the guard's precision and it currently describes a mixture.

**Fix sketch.** Give each its own term, the way call-shape already has one. The mechanical part is small — the checks are already separate calls with separate flags — but the finding-level provenance is not currently carried: `guard.Violation` has no field saying which check produced it, so `append` into a shared slice erases the origin at the call site. Either tag `Violation` with its source check and derive the reason terms from the tagged set, or keep per-check slices and merge only at render time.

[...]

PR #283, opened 2026-08-04
guard: gate the clean path on the findings, and pin the two unpinned buckets (#268 follow-up)

inth3shadows:
Follow-up to #270. Review of that PR found one defect visible from two sides.

## The defect

#268 gave file-scope, qualified and deps-go their own decision-log buckets by
recording which check fired in a `firedChecks` struct. The hook's clean-path
gate moved with it — from `len(violations)==0 && ...` on the merged slice to
`!fired.any()` on that struct. Those are not the same claim.

The three checks append INTO `violations` while separately setting their own
flag. Under `!fired.any()` a dropped flag assignment does not mislabel a log
line: it makes a findings slice holding real entries read as clean, and the
hook emits a `defer` where it used to ask. A silently dropped ask is the
guard's worst failure mode.

And that gate rested on bookkeeping nothing was pinning. On master, deleting
**both** `fired.Qualified = len(qv) > 0` and `fired.DepsGo = len(dv) > 0`
leaves `go test ./cmd/runecho-guard/` fully green — the two buckets #268 was
opened to make measurable were the two with no fixture behind them.
`fired.FileScope` was pinned, as #270's body claimed; the other two were not.

## Changes

**The gate** reads `len(violations) == 0 && !fired.anyNonViolation()` again, so
suppression by lost bookkeeping is structurally impossible rather than merely
tested-for. `any()` is replaced by `anyNonViolation()` — `Violations` is
deliberately absent, since the gate reads the slice directly — and its unit
test moves with it, gaining an assertion that folding `Violations` back in is
a regression.

**Hook fixtures for qualified and deps-go.** Both need worktree files the
harness could not express, so `hookCase` gains `files`. qualified needs a
`go.mod`, since it returns nil on an empty module path. deps-go needs a
`go.mod` plus a vendor tree, which keeps the fixture hermetic: a vendored build
resolves imports from `vendor/` and nothing else, so "the package resolved and
lacks this symbol" is a fact the fixture states rather than one inherited from
the machine.

**`expect_no_learn_symbols`** fails on a nil decision record instead of passing
vacuously. It is a safety pin — it asserts an ask did NOT train learned-allow —
so "nothing was logged" is the one outcome that cannot be read as agreement.
Its sibling `expect_log_reason` three lines below already fatals.

## Mutation score

Nine mutations added to `bench/hookmutate/mutations.json`. Each is caught by
exactly one fixture:

| Mutation | Caught by |
|---|---|
| `qualified/provenance` | `go-samerepo-missing-selector-asks` |
| `deps-go/provenance` | `go-vendored-dep-missing-selector-asks` |
| `qualified/repo-ir-lookup` | `go-samerepo-existing-selector-clean` |
| `qualified/samerepo-prefix-gate` | `go-external-import-is-not-this-checks-business-clean` |
| `deps-go/exported-membership` | `go-vendored-dep-existing-selector-clean` |
| `deps-go/unresolved-bail` | `go-unresolvable-dep-abstains-clean` |

Plus both flag-gates. The tenth, `gate/merged-slice`, **survives**, and is
recorded as a `known_gap` saying why rather than deleted: `fired.Violations` is
assigned from `len(violations)>0`, so in isolation the two spellings agree. It
is defense in depth against a SECOND mutation and is only observable compounded
with one. Compounded with `qualified/provenance`,
`go-samerepo-missing-selector-asks` fails with `expected an ask from the
qualified check, got a defer` — the ask dropped outright rather than
mislabelled, which is exactly what gating on the merged slice makes impossible.

## Known limitation

The deps-go fixtures exercise the **vendor** resolution path, while dogfooding
resolves through the module cache. Hermeticity was bought at the cost of
pinning the path users actually run. The provenance and rule mutations above
are indifferent to which path resolved the package, so the buckets are pinned
either way — but a module-cache regression in `depindex` would not fail this
corpus.

## Verification

- `go test ./... -race -cover` green; `cmd/runecho-guard` coverage 82.0% -> 83.5%
- `python3 bench/hookmutate/mutate.py --strict` exits 0 — no dead fixtures, all
  survivors carry a `known_gap`
- `go vet ./...`, `gofmt -l .` clean


inth3shadows:
Review addressed in d6d40db. No correctness defect was found in the gate change; all four findings were documentation or test-durability, and all four are fixed.

**`anyNonViolation`'s summary line was false for three of its seven fields.** It said it reports the checks whose findings do *not* land in `[]Violation` — but file-scope, qualified and deps-go all append into `violations`, as the same paragraph said two sentences later. Rewritten to say what the split actually is: four fields are the *only* record their check fired, and three are deliberate belt-and-braces redundancy with `len(violations)`.

**Both `firedChecks` tests enumerated the struct by hand** — the same unpinned-bookkeeping shape this PR exists to remove. They now walk it by reflection with `Violations` as a named exception, and `askReason` gains a twin test asserting every field maps to a **distinct** bucket. Verified against the exact scenario the review named:

```
$ # add `Contract bool` to firedChecks, wire it nowhere
dangling_test.go:154: anyNonViolation() ignores Contract — an ask with only that finding would be dropped
dangling_test.go:179: askReason returns "violations" for both Violations and Contract — Contract has
                      no bucket of its own, so its false-positive rate is unmeasurable
--- FAIL: TestFiredChecksAnyNonViolation
--- FAIL: TestAskReasonNamesEveryField
```

Before this change both stayed green.

**`go-external-import-is-not-this-checks-business-clean`'s desc was falsified by this PR's own mutation table.** It claimed the silence came from qualified abstaining rather than deps-go declining to run; `deps-go/flag-gate` kills the fixture, because default-on deps-go resolves `encoding/json` from GOROOT and flags `MarshalCanonical` legitimately. The desc now states the contingency and names the fix for the day deps-go graduates, since the failure (`expected no ask`) will not name it.

**Check counts.** The corpus header, `bench/hookmutate/README.md` and `mutate.py`'s docstring all still said five of six checks; with call-shape, qualified and deps-go it is eight of nine.

The same stale count appears in `.github/workflows/ci.yml:36`. Left untouched deliberately — it is only a comment, but the file is CI and gets its own change rather than riding along in a guard PR.

Re-verified: `go test ./... -race -cover` green (guard 83.5%), `mutate.py --strict` exit 0, `go vet` and `gofmt -l .` clean.

Issue #261, opened 2026-07-30
call-shape check is gated behind the store, but needs zero store data


inth3shadows:
The call-shape check (`RUNECHO_GUARD_CALLSHAPE=1`, shipped #243 slice 1 / v0.20.0) reads **only** the diff and the edited file's own lines:

```go
// cmd/runecho-guard/main.go:702
callShapes := callShapeMismatches(lang, fileLines, diffs[0], payload.ToolName, removedText)
```

No store, no snapshot, no symbol table. The declaration it compares against is parsed out of the file in front of it.

But it is wired at line 702 — **after** the `!res.OK` return at `cmd/runecho-guard/main.go:578-600`. So on an unenrolled tree (`res.NoRepo`) or an enrolled repo with no usable snapshot, the guard returns before call-shape ever runs, silently skipping a check that would have answered correctly.

Contracts already solved exactly this shape: `askContractOnly` is a pre-gate at 546 / 570 / 591 / 784 precisely because a contract binding "does not depend on the index". Call-shape has an even weaker dependency — it needs no store row at all, not even a repo binding.

**Priority: low.** The check is default-off, so today this only costs coverage for someone who has opted in and is editing outside an enrolled tree. Filing so it is not rediscovered.

**Fix sketch:** hoist call-shape ahead of the store gate, or give it an `askCallShapeOnly` sibling to `askContractOnly`. Prefer whichever keeps the "pay for your own store open" accounting in the surrounding comments honest — call-shape needs no store open at all, so it may be cheaper than contracts, not more expensive.

Not urgent, not blocking the #243 dogfood window.

inth3shadows:
Scoped this before implementing. The fix is not a move — it has a cost the contract pre-gate does not.

**Where things actually sit** (`cmd/runecho-guard/main.go`, v0.21.1):

| line | what |
|---|---|
| 573 | `lookupSymbolsFor` |
| 578-613 | the `!res.OK` gate — `askContractOnly` pre-gate, then return |
| 636 | `fileLines := readFileLines(filePath)` |
| 655 | `diffs := ...` |
| 702 | `callShapeMismatches(lang, fileLines, diffs[0], ...)` |

The premise holds: `callShapeMismatches` needs zero store data. But both of its inputs are built **after** the gate, so a pre-gate has to hoist `readFileLines` and the diff construction above line 573 — a file read plus a diff build on the ~12 ms budget for **every** edit in an unenrolled tree, which is the common case for a hook installed globally.

[...]

PR #284, opened 2026-08-04
docs: correct the corpus check count in CI and FINDINGS

inth3shadows:
Tail of #283, which moved the hook corpus from five of the guard's six checks to
eight of nine and corrected the count in the corpus header,
`bench/hookmutate/README.md` and `mutate.py`'s docstring.

Two copies were left behind deliberately, both outside that PR's scope:

- `.github/workflows/ci.yml:36` — `.github/` is a protected path here and earns
  its own change rather than riding along in a guard PR.
- `bench/FINDINGS.md:85` — not otherwise touched by #283.

Comment and prose only. No workflow behaviour changes: the single `ci.yml` line
is inside a `#` comment block, same indentation, and the file still parses (jobs
`test`/`docker`, five `test` steps ending in the mutation gate).

`FINDINGS.md` also gains an "or an on-disk module" clause. qualified and deps-go
are the first hook checks whose precondition lives on disk — a `go.mod`, a
vendor tree — rather than in an enrolled snapshot, which is why `hookCase`
needed a `files` field in #283.

Worth doing rather than shrugging at: the count is the corpus's own claim about
what it covers, and the entire argument for the mutation gate is that unmeasured
coverage is the hazard. A stale denominator is that same failure in prose.


inth3shadows:
Review found one issue, fixed in 99d34b5. It landed on the exact sentence this PR rewrites, which is a fair hit.

The line asserted of all eight hook-level checks that they "need edit old-vs-new text plus an enrolled snapshot store or an on-disk module." False for two of them. Verified against the fixture data rather than the prose:

| fixture set | `enroll`/`refs` | `files` |
|---|---|---|
| callshape | — | — |
| contract | yes | — |
| dangling | yes | — |
| deps_go | yes | yes |
| dropped_import | — | — |
| duplicate | yes | — |
| filescope | yes | — |
| qualified | yes | yes |

dropped-import reads the pre-edit file's own imports; call-shape deliberately answers on trees the store cannot resolve (#261/#266). The pre-existing text was already wrong for dropped-import — `hookcorpus_test.go:31-33` says outright that it "needs no snapshot enrollment at all" — and this PR had widened the disjunction for qualified/deps-go without fixing the universal underneath it.

Named the two exceptions rather than hedging with "most", matching this file's house style of stating gaps instead of implying them away.

Worth noting what the reviewer also checked and cleared, since it was the thing I most wanted a second opinion on: the **nine** denominator was counted from source (`askReason`'s eight fired checks plus `contract`, logged separately via `contractReason`), not taken from the prose being edited. The corpus directory holds exactly eight fixture files and `TestHookCorpus` globs `*.json`, so all eight are actually driven. `hookcorpus_test.go:48`'s surviving "six checks" is historically scoped and correct.

PR #285, opened 2026-08-04
fix(parser): ask heredoc eligibility a question `blanking` does not answer (#281, #282)

### LINKED ISSUE #281: parser(shell): bare `((x << y))` misread as a heredoc opener, swallows the rest of the file
## Summary

`maskShell`'s heredoc-opener guard (`internal/parser/shell.go:490`) is `blanking == 0`, but
`blanking` counts only the s/d/b/c/p frames — `frameSubshell` (`r`, pushed for a bare `(` at
`shell.go:466`) is *structural* and deliberately excluded.

So inside a bare arithmetic compound command `((...))` — two nested `frameSubshell` pushes,
**not** `$((...))` — `blanking` is still `0`, and the left-shift operator `<<` is read as a
heredoc redirect. The delimiter scan takes the following identifier as the terminator, and the
newline handler at `shell.go:523` then blanks every subsequent line looking for a terminator
line that never appears — swallowing the enclosing function's closing `}` and every function
defined after it.

## Reproduction

```sh
f() {
  ((x << y))
  echo after
}

g() {
  echo two
}
```

- **Actual:** `Parse()` returns `Functions=[f]`, `SymbolHashes={}` — `g` is dropped entirely,
  and `f` gets no body hash because its closing `}` was blanked while chasing terminator `y`.
- **Expected:** `Functions=[f, g]` with a non-empty body hash for both.

Verified by execution against the real `maskShell`/`Parse` (throwaway in-package `_test.go`,
run and removed).

## Root cause

`blanking` conflates two different questions:

1. *"Are we inside a quoted/expansion region where the bytes are literal?"* — correct for
   masking.
2. *"Is a `<<` here a heredoc redirect?"* — what line 490 actually needs.

`frameSubshell` answers (1) as "no" but must answer (2) as "no" too when the parens are an
arithmetic `((...))`. Same root cause as the sibling issue filed alongside this one, which is the mirror-image failure
(`frameCmdSub` *is* counted, and suppresses a heredoc that really is there).

## Suggested direction

Track `((` explicitly rather than pushing two `frameSubshell` frames — an arithmetic context
where `<<` is an operator, never a redirect. A narrower fix is to suppress heredoc-opener
detection when the top two stack frames are both `frameSubshell` pushed at adjacent offsets.

---
Found during a `/logic` sweep of `internal/parser/shell.go`; independently re-confirmed by an
execution-grounded verifier.

### LINKED ISSUE #282: parser(shell): heredoc inside `$(…)` never masked; heredoc body text extracted as a real function
## Summary

`maskShell` counts `frameCmdSub` (`c`, for `$(…)`) toward `blanking`, so both the heredoc-opener
guard (`internal/parser/shell.go:490`) and the heredoc-body blanker (`shell.go:523`) are
suppressed inside a command substitution.

That makes the very common `x=$(cat <<EOF … EOF)` pattern unparseable: no `hd` is ever pushed,
the heredoc body is never blanked, and the first literal `)` inside that body is read as the
closing paren of the `$(…)`. The command substitution ends early, masking of the remaining
content stops, and the raw heredoc body text is then scanned as real code.

## Reproduction

```sh
f() {
  x=$(cat <<EOF
)
fakefunc() {
EOF
)
  echo done
}
```

- **Actual:** `Parse()` returns `Functions=[f, fakefunc]` and hashes only `fakefunc` — which is
  heredoc *body text*, not a definition. The real function `f` gets no body hash at all,
  because `fakefunc`'s leaked `{` desyncs the brace-depth count.
- **Expected:** `Functions=[f]` only, with a body hash for `f` and no `fakefunc` entry.

Verified by execution against the real `maskShell`/`Parse` (throwaway in-package `_test.go`,
run and removed).

## Why it matters

This is the failure mode the masking pass exists to prevent — the docstring at `shell.go:18-21`
states that a `foo()` inside a heredoc "can therefore neither close a body early nor be mistaken
for a definition." Here it does both. A false symbol extracted from heredoc text is worse than a
missed one: it enters the snapshot as a real definition.

## Root cause

Mirror image of #281. Both stem from `blanking` being used to answer a question it does not
encode: *"can a heredoc start / continue here?"* `frameSubshell` is wrongly excluded there;
`frameCmdSub` is wrongly included here. A heredoc inside `$(…)` is legal shell and must still be
recognized and blanked.

## Suggested direction

Separate the two concerns — keep `blanking` for byte masking, and add a distinct predicate for
heredoc eligibility that is true at code level inside `frameCmdSub`, `frameSubshell` (non-
arithmetic), and `frameGroup`, and false inside `frameSingle`/`frameDouble`/`frameBacktick`/
`frameParam`.

---
Found during a `/logic` sweep of `internal/parser/shell.go`; independently re-confirmed by an
execution-grounded verifier.

inth3shadows:
Closes #281, closes #282.

## One conflation, two mirror-image bugs

`maskShell` keeps a single counter, `blanking`, incremented for the s/d/b/c/p
frames. It is used for two different questions:

1. **"Are these bytes literal?"** — masking. `blanking` is correct here.
2. **"Can a heredoc start / continue here?"** — `shell.go:490` and `:523`.
   `blanking` is wrong here, in *both* directions:

| context | `blanking` | heredoc legal? | before |
|---|---|---|---|
| `((x << y))` — two `frameSubshell`, structural, not counted | `0` | **no**, `<<` is a shift | opener fired → #281 |
| `$(cat <<EOF …)` — `frameCmdSub`, counted | `>0` | **yes**, legal shell | opener suppressed → #282 |

#281 swallows the enclosing function's closing brace and every function after it,
chasing a terminator that never comes. #282 is worse in kind: the unmasked heredoc
body is scanned as code and a definition that does not exist enters the snapshot as
a real symbol — the exact failure `shell.go:18-21` says the masking pass prevents.

Fixed as one thing, because one counter causes both and a partial fix invites the
next mirror image.

## Changes

- **`quoting`** counts only the frames whose *bytes* are literal (s/d/b/p). It
  differs from `blanking` by exactly `frameCmdSub`, and that difference is #282.
  **`blanking` is untouched** — masking behaviour must not move, and the existing
  masking tests are the regression net for that.
- **`inArith()`** identifies an arithmetic `((…))` or `$((…))` from frame kinds plus
  push offsets.
- **`heredocOK() = quoting == 0 && !inArith()`** replaces `blanking == 0` at both
  sites. Both must move together: fixing only the opener recognises a heredoc inside
  `$(…)` and then never blanks its body — strictly worse than before.

### Why a heuristic, and why not a `frameArith` frame kind

`((` is genuinely ambiguous — `((expr))` is arithmetic, `((cmd) | cmd2)` is a
subshell opening a subshell. Bash resolves it by trying to parse; a masker cannot.

A `frameArith` kind was **rejected**: it has to be popped by `))`, so a
misclassified `((echo a) | cat)` desyncs the stack on the first single `)` — and a
desynced masker swallowing the rest of the file *is* #281, reintroduced by its own
fix. `inArith` is instead a read-only predicate over frames pushed exactly as
before, so guessing wrong costs at most one missed heredoc.

`$((x << 2))` was previously safe only **by accident** — `$(` pushed a `frameCmdSub`
so `blanking > 0` suppressed the opener. Once `frameCmdSub` leaves `quoting` that
accident is gone, which is why `inArith` covers the `$((` spelling too.

## Verification

Seven table cases, each asserting **`Functions` AND `SymbolHashes`** — in both
issues the wrong-symbol and missing-hash halves are separate failures.

Mutation-scored; every mutation is caught by exactly the case that claims it:

| mutation | caught by |
|---|---|
| drop the arithmetic rule | both arithmetic cases |
| restore `frameCmdSub` to eligibility | both command-substitution cases |
| revert the body-blanker site only | dash/quoted-heredoc-in-`$(…)` |
| revert the opener site only | #281, #282, dash/quoted |
| drop the `$((` arm of `inArith` | arithmetic-expansion case |
| drop the offset-adjacency comparison | non-adjacent nested subshells |

**Two cases exist only because scoring demanded them**, which is the point of
scoring rather than arguing:

- `$((x << 2))` passes with the fix *removed* — the delimiter scan rejects a leading
  digit, so no heredoc is ever pushed and the case proves nothing. Now `<< shift`.
- Dropping the adjacency comparison survived every other case, so a
  `( (cat <<EOF …) )` case now pins that a non-adjacent nested subshell is not
  arithmetic and its heredoc is real.

Gates: `go test ./... -race -cover` green (`internal/parser` 82.8% → 82.9%),
`go vet`, `gofmt` clean.

## Limitations, recorded in the header rather than implied away

- A heredoc inside a backtick substitution is still not recognised: `` `…` `` content
  is masked wholesale without tracking inner structure, so there is nothing to hand a
  recognised heredoc to.
- `((` remains a heuristic. The bound that makes it acceptable is that a wrong guess
  costs a missed heredoc, never a desynced stack.


inth3shadows:
Review verdict was **not clean**, and it was right. Both findings were real regressions vs `master`, reproduced before fixing. Fixed in `1f5744e`.

I ran both constructed inputs against each branch rather than reasoning about them:

| case | master | first fix | now |
|---|---|---|---|
| `((cat <<EOF … ) \| tr a-z A-Z)` | `[f g]` | `[f fake_adj g]` | `[f g]` |
| `x=$(( (1 << bit) ))` | `[f g]` | `[f]` | `[f g]` |
| `(( m = (1 << bit) ))` | `[f]` | `[f]` | `[f g]` |

The third was broken on **both** branches — a pre-existing bug the same fix closes.

## The cost bound in my own comment was wrong

I wrote that "the cost of guessing wrong is at most one missed heredoc", and flagged in my own summary that this was an argument no test pinned. The review found the case where it fails, and the asymmetry is the whole point:

- inside `$((…))` — `frameCmdSub` is a **blanking** frame, contents masked wholesale. A missed heredoc there really is just missed.
- inside `((…))` — `frameSubshell` is **structural**, contents kept as code. A missed heredoc is *not* a miss: its body is scanned and `fake_adj() {` becomes a definition that does not exist. That is #282's failure mode reached from the other side, and worse than a miss.

`arithShift` now takes that asymmetry as its shape: suppress freely when `blanking > 0`; in the structural case only when a `))` closes on the **same line** as the operator. Real arithmetic closes on its operator's line — `((cat <<EOF` does not.

`inArithNest` walks the run of subshell frames instead of checking the top pair, which is finding 2 and also closes the pre-existing bare-`((` spelling.

## Re-scored

| mutation | caught by |
|---|---|
| neuter the arithmetic rule | all three arithmetic cases |
| same-line `))` confirmation always succeeds | adjacent-paren pipeline (finding 1) |
| check only the top pair | nested subexpression (finding 2) |
| drop the `blanking > 0` short-circuit | multi-line `$((…))` |
| drop the `$((` arm | both `$((` cases |
| restore `frameCmdSub` to eligibility | both command-substitution cases |

The multi-line `$((…))` case is new and pins a **no-regression-vs-master** property: `$(` used to make `blanking > 0` there, so demanding the same-line `))` in that path too would swallow `g` where master does not.

## One survivor, recorded rather than covered

Offset adjacency now survives mutation — the same-line `))` confirmation independently rejects every non-adjacent case reachable in practice, and in the `blanking > 0` path a wrong answer is unobservable because the frame masks its contents anyway. So it is redundancy, not load-bearing. I noted that in `inArithNest` instead of inventing a fixture for it; a test claiming to cover it would be the vacuous-coverage antipattern this package's tests exist to avoid.

Header limitation updated too: multi-line **bare** `((` arithmetic still mis-opens a heredoc, unchanged from before this branch.

`go test ./... -race -cover` green, `internal/parser` 82.8% → 83.1%.

inth3shadows:
Closing unmerged. Three rounds of review found real regressions, each one in the heuristic layer, and each round's fix introduced the next round's defect:

1. `inArith` suppressed heredocs in a structural frame (phantom symbol), and checked only the top stack pair (whole-file loss on `$(( (1 << bit) ))`).
2. The `))`-on-this-line confirmation could not tell a subshell pipeline whose own closer is `))` from arithmetic — the same phantom-symbol class, on a one-character variation of its own fixture.
3. `arithShift`'s safety invariant is **false**. I wrote that suppressing inside `$((…))` is safe "because the frame masks its contents wholesale and a missed heredoc emits nothing." Masking the *bytes* does not stop the frame stack walking the body: an unmatched `{` in it prevents the `$(` from ever popping, and everything to EOF is swallowed.

The common factor is a length-preserving byte masker guessing at shell grammar from a frame stack. `((` is ambiguous in a way bash resolves by *attempting to parse*, which this engine structurally cannot do. Patching further pays down a design that should be deleted.

**The premise that justified the masker has changed.** `shell.go`'s header argues "a full tree-sitter-bash grammar is heavier than the symbol model warrants" — written as if adding a grammar meant adding a dependency. It does not. Verified against the module cache:

```
gotreesitter@v0.47.1/grammars/
  bash_scanner.go   z_subset_blob_embed_bash.go   z_subset_registry_register_bash.go
```

The bash grammar already ships in the dependency this repo uses for Python, JS, TS, TSX, Rust and Ruby, behind the same `grammar_subset_*` build-tag mechanism.

So #281 and #282 stay **open**, to be fixed by replacing the masker rather than repairing it. `master` keeps both bugs, which are known and bounded, instead of trading them for a heuristic that fails review.

Everything from this branch that is worth keeping is carried forward as the acceptance criteria for that work: the table cases for #281/#282, and the executed reproductions from all three review rounds (`((cat <<EOF) | (tr a-z A-Z))`, `x=$(let y=1<<n)`, `x="$(cat <<EOF …)"`, stacked `<<A <<B`, unterminated-at-EOF, `$( ((cat <<EOF …`). An AST should pass all of them with no special handling — that is the test.

Plan: `~/.claude/plans/runecho-shell-tree-sitter-bash.md`.

PR #287, opened 2026-08-04
parser(shell): replace maskShell with tree-sitter-bash (#281, #282)

### LINKED ISSUE #281: parser(shell): bare `((x << y))` misread as a heredoc opener, swallows the rest of the file

### LINKED ISSUE #282: parser(shell): heredoc inside `$(…)` never masked; heredoc body text extracted as a real function

inth3shadows:
## Summary

Replaces `maskShell` (the shell parser's length-preserving byte masker) with a
tree-sitter-bash AST walk, fixing #281 and #282.

Three review rounds of patching the masker's frame-stack heuristic (PR #285,
closed unmerged) each fixed one heredoc/arithmetic ambiguity and introduced
another — `((` is ambiguous in a way bash resolves by attempting to parse,
which a length-preserving masker structurally cannot do. This replaces the
masker entirely with the same vendored tree-sitter grammar already used for
Python/JS/TS/Rust/Ruby, behind the same `grammar_subset_*` build-tag
mechanism.

Along the way, found a real gap in the grammar itself: stacked heredocs on
one line (`<<A <<B`) mis-parse, and error recovery can fabricate a bogus
`function_definition` out of what is actually heredoc body text (confirmed
by direct AST inspection). Worked around with a narrow, heredoc-only-aware
fallback — no quote/paren/subshell frame stack, so it can't reintroduce the
`((` ambiguity above — that cross-checks the AST result by intersection
whenever the parse errors, rather than a plain union (shell's failure mode on
error is over-reporting, not just under-reporting, so union alone isn't
safe).

Adds `grammar_subset_bash` to both shipping channels (`install.sh`,
`.goreleaser.yaml`) so the parser isn't inert in the binaries this project
actually ships — see #229/#264 for the precedent (Rust/Ruby shipped inert for
the same reason).

## Changes

- `internal/parser/shell.go`: `maskShell` and its frame stack deleted; new
  AST walk + heredoc-aware fallback for the error path.
- `internal/parser/shell_treesitter_acceptance_test.go`: the #281/#282
  reproductions as acceptance tests (now passing).
- `internal/parser/shell_fuzz_test.go`: rewritten to fuzz the new engine's
  actual invariant (never panics) instead of the deleted masker's
  length-preservation invariant.
- `install.sh`, `.goreleaser.yaml`: `grammar_subset_bash` added to every
  build-tag list.

## Test plan

- [x] `go test ./... -race -cover` green (whole repo)
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `bash -n install.sh` valid
- [x] Full `shell_test.go` suite (the pre-existing contract) passes unchanged
- [x] `TestInstallShipsEveryGrammarTag` passes (both shipping channels + the
      Dockerfile check)
- [x] Built the actual shipped binary via `install.sh` and verified it
      indexes a real `.sh` file correctly — extended-charset names, the
      `((x << y))` #281 pattern, all with body hashes
- [x] Binary size impact measured: +~220 KB (21.1 MB → 21.3 MB, ~1%)
- [x] Index time on a 91-file real-world shell corpus measured: ~4.7s →
      ~5.5s average (background-only cost, not hook-blocking)

Closes #281. Closes #282.


inth3shadows:
Ran \`/code-review\` against this PR — 4 findings, all confirmed by direct execution and fixed in b0d7e35:

1. **Total-fallback path was quote-blind**: a def-shaped line inside a multi-line string got extracted as a real function when the AST was entirely unavailable. Fixed by giving the shared fallback mask real quote/comment/backtick/param-expansion tracking.
2. **Partial-parse cross-check could silently drop a real function**: the old name-intersection filter used a line-anchored regex that can't see a definition after `;` on the same line as another. Replaced with a per-node byte-position check during the AST walk itself — trust a `function_definition` unless its own start lands in the shared mask's blanked region.
3. **Zero-AST-functions fallback branch lost all body hashes file-wide**, not just for the specific function that needed it. Reintroduced a minimal brace/paren depth-counter scoped to the now quote-safe masked buffer.
4. **`TestASTGrammarsLoad` didn't check `bashLanguage()`** — the same silent-nil-fallback gap its own comment describes for Rust/Ruby/JS/TS. Added the missing check.

One thing the fix for #2 surfaced: gating the byte-position filter on `HasError()` turned out to be load-bearing, not optional — the shared mask's heredoc detector has the same `<<`-vs-arithmetic blind spot as the deleted `maskShell` (by construction; it deliberately doesn't track parens), so applying it unconditionally reintroduced #281 one layer up. Caught by re-running the acceptance suite, documented in the commit and in code.

Full repo suite (`go test ./... -race -cover`), `go vet`, and `gofmt` all green after the fixes; shell tests pass under both default and shipped (`grammar_subset_bash`) build tags.

PR #288, opened 2026-08-04
guard: fix 7 extract.go false positives/negatives (#274-#280)

### LINKED ISSUE #274: guard: isFStringPrefix returns true for invalid f-string prefixes (bf, fb, ff)
## Defect

`isFStringPrefix` at `internal/guard/extract.go:1497` scans the run of prefix letters before a quote and returns `true` if ANY character is `f` or `F`. Valid Python 2-letter prefix combinations containing `f` that are NOT f-strings (`bf"..."`, `fb"..."`, `ff"..."`) are misidentified as f-strings, causing the string-literal scanner to attempt f-string interpolation parsing on non-f-string bytes/format literals.

## Reproduce

`isFStringPrefix([]byte("bf\""), 2)` returns true — should be false. `bf` is a valid Python bytes-literal-with-format-spec prefix, not an f-string.

## Current behavior

Returns `true` for any 2-letter prefix containing `f` — `bf`, `fb`, `ff`, and all case variants.

## Expected behavior

Should only return `true` for valid Python f-string prefixes: bare `f`/`F`, or `rf`/`fr`/`RF`/`FR`/`rF`/`fR`/`Rf`/`Fr` (raw f-string). `bf`/`fb`/`ff` are not valid f-string prefix combinations.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #275: guard: parseJSBindingTarget does not handle array destructuring
## Defect

`parseJSBindingTarget` at `internal/guard/extract.go:359` only handles object destructuring (`{a, b}`) and bare identifiers. Array destructuring (`[a, b]`) falls through to the bare-identifier branch where `reIdent.MatchString("[a, b]")` fails (anchored with `^`, `[` doesn't match), returning `nil`.

## Reproduce

For `const [a, b] = require('m')`, `parseJSBindingTarget("[a, b]")` returns `nil`. Expected: `["a", "b"]`.

## Current behavior

Zero names extracted from array destructuring imports. The imported names are not added to the known set, so genuine calls to `a(...)` or `b(...)` later in the file may be falsely flagged as hallucinations.

## Expected behavior

Array destructuring targets should be extracted — strip `[`/`]` brackets, split on commas, match each identifier.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #276: guard: parseJSBindingTarget nested destructuring corrupted by cutset trim
## Defect

`parseJSBindingTarget` at `internal/guard/extract.go:363` uses `strings.Trim(s, "{}")` to strip outer braces from object destructuring. `strings.Trim` operates on a *cutset*, not a matched pair — it strips all leading/trailing characters in the set `{}`. For nested destructuring like `{a: {b, c}}`, this produces `a: {b, c` (only outermost braces removed), then `strings.Split(", ")` yields `["a: {b", " c"]`. The item `{b` passes `reIdent.MatchString` (unanchored, matches `b`) and is appended as `"{b"` — not a valid identifier.

## Reproduce

For `const {a: {b, c}} = require('m')`, `parseJSBindingTarget("{a: {b, c}}")` should return `["c"]` (the bound names after rename), but produces corrupted identifiers.

## Current behavior

Nested object destructuring produces corrupted identifier strings.

## Expected behavior

Either handle arbitrary nesting depth, or at minimum produce clean identifiers for the bound names at each level.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #277: guard: appendConstRefs skips dict keys as false annotation targets
## Defect

`appendConstRefs` at `internal/guard/extract.go:885` checks whether `rest` (the text after a matched `UPPER_SNAKE` identifier) starts with `:` or `=`. If `rest` starts with `:`, the identifier is skipped as an "annotation/assignment target." But in a dict literal like `{MAX: 5}`, `MAX` is followed by `:` — it's a dict key (a genuine *use* of the constant, not a definition), and the guard should check that `MAX` resolves. Instead it's silently skipped.

## Reproduce

In guard-processed code containing `result = {MAX: 5}`, `appendConstRefs` matches `MAX`, sees `rest=": 5}"` starting with `:`, and skips it. `MAX` is never checked as a reference.

## Current behavior

`UPPER_SNAKE` identifiers used as dict keys are silently skipped and never validated as references.

## Expected behavior

Distinguish between annotation targets (`MAX: int = 5` — a type annotation on a definition) and dict keys (`{MAX: 5}` — a reference). The `:` test is too broad.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #278: guard: appendConstRefs treats tuple assignment targets as references
## Defect

`appendConstRefs` at `internal/guard/extract.go:885` skips identifiers followed by `:` or `=`. In tuple/multiple assignment like `MAX, OTHER = 5, 10`, after matching `MAX` the rest is `, OTHER = 5, 10` — which starts neither with `:` nor `=`. So `MAX` is added as a *reference*, but it's actually being *defined* (assigned to). Calls to `MAX` later in the file would be flagged as "unresolved" because the guard thinks `MAX` was used-but-not-defined.

## Reproduce

In guard-processed code containing `MAX, OTHER = 5, 10`, `appendConstRefs` adds `MAX` to refs (thinks it's a use), but `MAX` is being defined here. This is a false positive.

## Current behavior

The first identifier in a tuple assignment is treated as a reference, not a definition. This can cause false-positive "unresolved reference" findings.

## Expected behavior

Detect tuple/multiple assignment and treat all names on the LHS as definitions, not references. The check should consider `,` as a continuation of the assignment target list, not fall through when neither `:` nor `=` immediately follows.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #279: guard: defNames uses FindStringSubmatch, only captures first definition per line
## Defect

`defNames` at `internal/guard/extract.go:1599` calls `re.FindStringSubmatch(text)` on definition regexes. This only captures the *first* match on a line. For lines with multiple definitions — `var a, b = 1, 2`, `const x, y = foo()`, `let a, b = bar()` — only the first name is added to the known set. The second name is missing from `known`, so a later reference to it may be falsely flagged as a hallucination.

## Reproduce

For `var a, b = 1, 2`, `defNames` only returns `{"a"}`, not `{"a", "b"}`. A later reference to `b()` is falsely flagged as a hallucination.

## Current behavior

Only the first definition per line is captured.

## Expected behavior

Use `FindAllStringSubmatch` to capture all definitions on a line.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

### LINKED ISSUE #280: guard: isImportLine (JS) requires literal spaces around import/from keywords
## Defect

`isImportLine` at `internal/guard/extract.go:1627,1633` uses `strings.HasPrefix(t, "import ")` and `strings.Contains(t, " from ")` — both require a literal space character after/before the keyword. Minified or space-less syntax like `import{a}from"m"` and `export{a}from"m"` is not recognized as an import line.

## Reproduce

`isImportLine(LangJS, "import{a}from\"m\"")` returns false — should be true.

## Current behavior

Space-less JS import/export syntax is not recognized. References on those lines aren't skipped, so imported bindings may be falsely checked as unresolved calls.

## Expected behavior

Detect import/export syntax without requiring spaces. Check for `import{` and `}from` variants, or use a regex that allows optional whitespace.

## Severity

Low — minified code is rare in git diffs. A codebase that runs a minifier or bundler before committing (e.g., vendored deps) would be affected.

## Source

Found by quarry gap-finder, execution-verified. Commit: `3869aa2`.

inth3shadows:
## Summary

Fixes all 7 issues in the guard bug-fix tail from the quarry gap-finder batch (commit 3869aa2). All same file (`internal/guard/extract.go`), independently scoped, batched into one PR.

Each was execution-verified before fixing — several of the issues' own reproduction examples didn't actually trigger the described code path (e.g. #279's JS/Python examples don't match the relevant regexes at all due to shape/charset requirements; #276's claimed "corrupted identifiers" actually silently drops a name cleanly instead). Fixed the real, confirmed defect in each case rather than the issue text literally, and used realistic repro inputs in the tests.

- **#274** `isFStringPrefix`: `bf`/`fb`/`ff` wrongly read as f-string prefixes (a "contains f" check instead of an exact valid-combination check).
- **#275 / #276** `parseJSBindingTarget`: array destructuring returned `nil`; nested object destructuring silently dropped names (`strings.Trim(s, "{}")` is a cutset trim, not brace-pair-aware, so it only strips the outermost pair). Fixed by delegating to `jsBindingTargets` — the bracket-depth-aware parser `JSDeclaredNames` already uses for the identical LHS-pattern shape — rather than reimplementing bracket matching a second time.
- **#277 / #278** `appendConstRefs`: a dict key (`{MAX: 5}`) was skipped as if it were a type annotation; a tuple-assignment target (`MAX, OTHER = 5, 10`) was added as a reference instead of recognized as a definition. Fixed with a statement-start position gate (distinguishes `MAX: 5` at line-start from `{MAX: 5}` inside a literal) plus a dedicated tuple-target-span regex covering every name in the list, not just the first.
- **#279** `defNames`: only captured the first name on a multi-declarator line. The issue's suggested fix (switch to `FindAllStringSubmatch`) turns out to be a no-op — these regexes are all `^`-anchored, and RE2's `^` only ever matches true string-start, never a resumed-search position, so `FindAll` can't find a second match on a single-line string regardless of which Find variant is used. Real fix: unanchored continuation regexes (`reJSVarDefCont`, `rePyConstDefChain`) that `FindAll` can walk across the rest of the line for each additional declarator.
- **#280** `isImportLine`: minified/bundled `import{a}from"m"` syntax wasn't recognized, only the spaced form.

## Test plan

- [x] One permanent regression test per fix in `extract_test.go`, plus guards for adjacent correct behavior each fix must not break (annotation-target skip, call-argument-list references, `==` not misread as chained assignment, dynamic `import(x)` not misread as an import statement)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go test ./internal/guard/...` green
- [x] `go test ./... -race -cover` green (whole repo)

Closes #274. Closes #275. Closes #276. Closes #277. Closes #278. Closes #279. Closes #280.


inth3shadows:
Ran \`/code-review\` — 4 findings, all real. Fixed 3, filed the 4th as a tracked follow-up (#289):

1. **Real regression**: \`appendConstRefs\` never consulted \`defNames\`' \`defs\` map, so a chained-assignment target \`defNames\` correctly recognized (\`OTHER_B\` in \`MAX_A = OTHER_B = 5\`) was still added as a plain reference — my earlier fix to \`defNames\` never reached the function that actually decides ref-vs-definition for SCREAMING_SNAKE names. Fixed by threading \`defs\` through (mirroring how the call-ref path already does this).
2. **The regex chain-detector I wrote for #279 had its own bug**: it could supply at most one extra name per chained assignment, because FindAll's non-overlapping match consumes the trailing `=` guard the next match needs. A 3-way-or-longer chain (`A = B = C = 5`) silently dropped everything past the second name. Replaced with a split-based approach (`pyChainedAssignTargets`) that has no such limit — verified up to 4-way.
3. **Minified-export detection was too literal**: required the exact substring `"}from"`, missing a partially-spaced mix like `export{a} from "m"`. Replaced with a single word-bounded regex.
4. **Not fixed here** — `appendConstRefs`' dict-key-vs-annotation check is line-local, so it still misses the common multi-line dict-literal form (a key on its own line still looks statement-start). Not a regression — the deleted `maskShell` had the same blind spot — but it undercuts what #277's fix claims to cover. Fixing it properly needs cross-line brace-depth tracking (Python dicts nest arbitrarily, unlike the Go-interface boolean this codebase already uses for a similar problem), which is real scope beyond this batch. Filed as #289 rather than rushed.

All fixes verified by direct execution (not just re-reading the diff), including the 3-way/4-way chain cases and the partial-spacing import variants. Full repo suite (\`go test ./... -race -cover\`), \`go vet\`, \`gofmt\` all green.

PR #308, opened 2026-08-06
guard: adjudicate resolution verdicts with the Go compiler

inth3shadows:
## Why

Every existing way this repo finds guard defects derives "correct" from RunEcho itself — the testdata corpora, the hook fixtures in `bench/hookmutate`, the fuzz seeds, and `fpaudit` (oracle = git history). The one exception, `callshapecheck_differential_test.go` adjudicated by CPython's `ast`, is also the one that produced hard numbers at scale.

This generalises that exception to the guard's *primary* claim — "identifier X does not resolve" — with `go build` as the adjudicator.

Both directions are machine-proven. No human judgement, no labelled corpus:

- a package that compiles proves every identifier in it resolves, so a guard flag there is a **proven false positive**;
- rewriting a use site to a name absent from the package makes the compiler say `undefined:`, so guard silence there is a **proven false negative**.

The false-negative direction is new. A corpus only contains violations someone already thought of, and `fpaudit` only sees decisions the guard already emitted — nothing until now could measure a miss.

## What it measured

Stratified by the shape of the reference. 240 adjudicated mutations on this repo, pattern reproduced on `golang.org/x/text`:

| shape | proven unresolved | caught | missed |
|---|---|---|---|
| `Foo()` bare exported | 15 | 15 (Run) | **0** |
| `pkg.Foo()` qualified | 60 | 58 (dep + same-module) | 2 |
| `foo()` bare unexported | 29 | 0 | 29 — documented policy |
| `v.Method()` on a value | 60 | 1 | 59 |

**Zero proven false positives over 468,675 lines of foreign Go** (`x/text`), and zero on this repo. The foreign number is the one that counts: runecho's own Go is the code the guard was tuned against.

## The receiver-method check

`recvmethod.go` closes the one slice of `v.Method()` that needs no type inference: inside `func (r *T)`, the receiver's type is written on the line that opens the method. Five gates keep it precision-first, the last being an embedding backstop (fire only on a name the repo has never seen in any form).

Measured: **0 false positives on both corpora**, and it does fire — not vacuous. But a direct population probe over this repo's 2,555 selector sites puts its ceiling at **0.5%**: base is a declared receiver var 4.9% → survives the ambiguity/rebinding gates 3.2% → member is exported 0.5%. Sibling-method calls are usually to *unexported* helpers, which the IR does not index.

It is **not wired into the hook** — it runs in the differential only, until the follow-up below lands and makes it worth more than 0.5%.

## The result that matters

That ceiling and the `foo()` blind spot are the same root cause: `internal/parser/go.go` gates unexported symbols out of the IR (four `ast.IsExported` calls). That one decision produces both the 2,118-site bare-unexported gap and the receiver check's cap. They move together or not at all. Follow-up PR.

## Fidelity note

`FoldInFileDefs` moves the hook's in-file definition fold from `cmd/runecho-guard` into the guard package. A harness that *mirrored* it would drift, and a drifting known set turns every false-positive number into fiction. One implementation, both callers.

## Running it

```
RUNECHO_ORACLE_CORPUS=<module> RUNECHO_ORACLE_MUTATIONS=240 \
  go test ./internal/guard -run TestGoResolve -v
```

Defaults to this repo. Phase 2 skips under `-short`. Caps and oracle-discard counts are always logged — a truncated run must never read as full coverage.

## Verification

`go build ./...` and `go test ./... -race -cover` green; `gofmt -l .` clean under the go.mod-pinned 1.25 toolchain.


inth3shadows:
Follow-up review found a methodological blind spot in this harness, corrected in #310: `runGuardOnFile` passed the whole file as **both** the added text and the fold source, so the known set always already contained every declaration in the file — a condition neither the hunk-scoped hook path nor the pre-commit path ever satisfies.

The "zero proven false positives over 468k lines" result in this PR body therefore holds only for the whole-file Write posture. #310 adds `write-new` (no fold) and `edit-hunk` (one function as the hunk, the rest as the pre-edit fold) postures, and re-establishes the zero across all three.

Merging this first as-is; #310 carries the correction.

Issue #301, opened 2026-08-06
ir: RUNECHO_HOME does not isolate git hooks — repo add rewrites shared .bare/hooks for every worktree


inth3shadows:
`RUNECHO_HOME` redirects the *store* (`decisions.jsonl`, `history.db`). It does
not redirect git hooks. `runecho-ir repo add` calls `installHooks`
unconditionally, and `installHooks` writes into
`$(git rev-parse --git-common-dir)/hooks` — which in a bare-worktree layout is
`.bare/hooks`, **shared by every worktree of the repo**. The hook bodies point at
whichever binary invoked `repo add`.

## How it bit

During review of #299, a scratch-built binary was run with `RUNECHO_HOME` set to
a sandbox, on the assumption that isolated the machine. It repointed this repo's
shared hooks to the scratch build for every worktree. `RUNECHO_HOME` reads like
protection and is not.

The failure is silent in the direction that matters: the hooks keep working, so
nothing reports an error — they are simply running a different binary than the
installed one, and the three reindex hooks are what every guard answer is
computed from (see the hook table in `CLAUDE.md`).

## Proposed fix

Either would do:

- a `--no-hooks` flag on `repo add`, so tooling and subagents have a documented
  way to enroll a repo without touching shared state; or
- skip hook installation when the invoking binary is not the installed one
  (resolve `os.Executable()` against the install path) — safer default, no flag
  to remember.

Also worth doing regardless: state in the docs that `RUNECHO_HOME` scopes the
store only, since the name invites the broader reading.

## Related

Subagents asked to review this repo need an explicit no-write constraint naming
`repo add` / `reindex` / `install`; "use RUNECHO_HOME" is not that constraint.

inth3shadows:
Retagged priority-2 → priority-1 via /prioritize.

Reason: this is the same silent-corruption failure class as #228 (a scratch/dev binary silently repointing shared hook state, degrading every consumer with no error) — #228 was resolved via a dedicated fix (PR #232) rather than left at a lower tier. Here the corrupted state is `.bare/hooks` via `repo add` from a scratch binary, affecting every sibling worktree. Given this workflow runs `claudew`/worktrees routinely, the trigger condition isn't rare, and the failure is undetectable until something downstream (reindex, guard) quietly stops working.

inth3shadows:
## Fix options (not selecting one yet)

[...]

Issue #294, opened 2026-08-05
guard: PyDeclaredNames/LocallyBoundNames have their own unseeded bracket trackers — same false-negative class as #289 for non-const bindings


inth3shadows:
## Observation

#289 (and its follow-ups in PR #290) fixed the "name at statement-start inside an existing multi-line literal misreads as a definition" false negative specifically for SCREAMING_SNAKE constant dict keys (`appendConstRefs`/`defNamesInContext`/`ExtractDefs`'s Python branch).

`PyDeclaredNames` and `LocallyBoundNames` (used by `pyFileScope`, and by the additive check's own known-set folding in `validate.go`) have their OWN independent, unseeded bracket-depth trackers for the analogous problem in non-constant bindings — e.g. a kwarg-style name (`timeout=30,`) sitting on its own line inside a pre-existing multi-line call whose opener is unchanged context above the hunk.

## Verified scenario

A hunk consisting of `    timeout=30,` sitting inside a pre-existing multi-line call whose opener is unchanged context above the hunk. `PyDeclaredNames`' unseeded tracker still misreads this as a kwarg-style binding and folds `timeout` into scope, silently suppressing a genuine unresolved `timeout(...)` reference elsewhere in the same diff — same false-negative shape #289 fixed for constants, unaddressed for this sibling binder.

## Why this isn't blocking #289's merge

This is a pre-existing limitation these functions already had, not something #289/#290 broke — #289 was scoped specifically to the SCREAMING_SNAKE dict-key case. This issue is scope for a broader follow-on audit of every extractor with its own cross-line bracket state, not a regression from #290.

## Fix sketch

Whatever seeding mechanism proves durable for `pyBraceDepth` (see #289/#290) should be generalized and threaded into `PyDeclaredNames`/`LocallyBoundNames`'s own trackers too, rather than each extractor inventing its own.

## Source

Found via code-review on PR #290 (round 3, line-by-line finder).

inth3shadows:
Round-3 code review on PR #297 confirmed this with an executable repro and scoped it precisely.

`PyDeclaredNames` and `PyParamNames` count brackets with `pyBracketDelta` over the **code** scan, where f-string interpolations are still intact. PR #297 fixed the three *dict-depth* consumers (`extractRefs`, `PyBraceDepthBefore`, `pyBraceDepthSeedFor`) to read `braceScan` instead, but these two were not part of that set — so they keep the pre-#291 exposure.

Repro (verified on PR #297's head):

```python
[...]

Issue #289, opened 2026-08-04
guard: appendConstRefs dict-key detection is line-local, misses multi-line dict literals


inth3shadows:
## Defect

\`appendConstRefs\` (\`internal/guard/extract.go\`) decides whether a SCREAMING_SNAKE name is a dict key (reference) or an annotation/definition target using a statement-start position check: is the name preceded by only whitespace on its own line? That correctly distinguishes \`{MAX: 5}\` (MAX preceded by \`{\`, a reference) from \`MAX: int = 5\` (MAX at true line start, a definition) — but only within a single physical line.

A real multi-line dict:

\`\`\`python
result = {
    MAX_VALUE: 5,
}
\`\`\`

puts \`MAX_VALUE\` on its own line, preceded only by whitespace *within that line* — indistinguishable, at the single-line \`scan\` this function operates on, from a genuine statement-start annotation. It is silently treated as a definition and skipped, so a hallucinated/undefined \`MAX_VALUE\` used as a multi-line dict key would not be flagged.

## Verified

\`\`\`go
appendConstRefs(nil, map[string]struct{}{}, "    MAX_VALUE: 5,", 2, nil, nil)
// returns [] — no reference recorded
\`\`\`

## Why this isn't a regression

The deleted \`maskShell\`-era code had the identical blind spot (a single-line masked-scan model with no cross-line brace state). #277 fixed the single-line inline case (\`{MAX: 5}\`); this is the pre-existing gap the fix does not reach.

## Fix sketch

\`ExtractRefs\` already tracks cross-line state for an analogous problem — \`inIface\` (Go interface bodies) is a boolean carried across lines, reset on a diff-hunk gap. Its own comment notes it gets away with a boolean because "interface bodies hold no nested braces in practice." Python dict literals don't have that property (they nest arbitrarily), so the equivalent tracker here would need real brace-depth counting, not just a boolean — genuinely more scope than a boolean flag, which is why this was filed separately rather than folded into the #274-#280 batch.

## Source

Found via code-review on PR #288 (the #274-#280 batch), which added the position-based dict-key/annotation distinction this issue is about the limits of.

PR #334, opened 2026-08-11
guard: RUNECHO_HOME hook-install warning (#301) + converge Python bracket trackers (#294)

inth3shadows:
## Summary

Two independent fixes from the parallel-gated build plan
(`~/.claude/plans/runecho-parallel-gated-build-plan-2026-08-11.md`), landed as
separate commits.

**#301 — `installHooks` warns on a non-installed binary** (`31231f9`)

`repo add`/`install` write hook bodies pointing at `os.Executable()`, shared
by every worktree via the common git dir. `RUNECHO_HOME` scopes the store
only, not hooks, so a scratch build could silently repoint a repo's shared
hooks with no signal at all. `installHooks` now warns (stderr, non-blocking)
when the running binary differs from what `runecho-ir` resolves to on PATH.
Also corrects `TECHNICAL.md`'s `RUNECHO_HOME` row, which called it an
"isolation / testing seam."

**#294 — converge Python's unseeded bracket trackers** (`9a4a493`)

`PyDeclaredNames`/`PyParamNames` each ran their own local bracket-depth
counter starting unconditionally at 0, unlike the `pyBraceDepth` tracker
#289/#291 already fixed. A hunk beginning mid-signature or mid-call (opener
in unchanged context above the hunk) was scanned as if no bracket were
open — two of these call sites are on the default `guard.Run` path, not
behind a flag. Converges onto two new shared, seeded trackers
(`PyBracketDepthBefore`/`PyDefSigDepthBefore`) instead of a fourth hand-kept
copy, threaded through every real hunk-scoped call site.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l` all clean
- [x] Full test suite (`go test ./...`) green across all 16 packages
- [x] New regression tests for #301: mismatch warns, match is silent,
      unresolvable-path fallback, PATH-lookup-failure is silent
- [x] New regression tests for #294: unit tests for both new `*Before`
      seed functions, seeded-vs-unseeded tests for
      `PyDeclaredNames`/`PyParamNames`/`LocallyBoundNames` (each with an
      explicit unseeded control proving the bug reproduces without the fix),
      and one end-to-end `Run()`-level test reproducing #294's own
      "Verified scenario" directly


inth3shadows:
Ran \`/code-review\` against this PR — one real correctness bug found and fixed (18ff502), three lower-severity findings left open for a decision.

**Fixed:** \`PyParamNames\`' \`sigDepth\` advance tracks all of \`()/[]/{}\`, but the seed I gave it (\`PyDefSigDepthBefore\`, shared with \`LocallyBoundNames\`) tracked parens only. A default value spanning a multi-line list/dict/set desynced the two — verified repro: a hunk starting inside such a literal silently dropped every parameter after it, the exact false-negative class this PR exists to close. Gave \`PyParamNames\` its own seed family using its own counting rule, and guarded a second-order issue (\`pyParseParamList\` assumes its input starts at the signature's own top level — only true at seed depth exactly 1) so a deeper seed falls back to the safe pre-#294 miss rather than risking a corrupted parse.

**Open, not addressed in this PR:**
1. \`Run()\` now does 4 independent file reads/re-scans per Python \`FileDiff\` on the pre-commit path (one per seed function). Real, but pre-commit-only, not the hot hook path.
2. \`hookBlockIndices\` is now called 4x with identical arguments per hook invocation (once per \`*ByLine\` seed builder) — this one IS on the ~12ms hook budget.
3. Structural observation: this PR's seed scaffolding (ReadFile + size cap + prefix-array + clamping closure) is now hand-copied 4 times, differing only in the per-line delta function — a case for a shared "seeded depth by delta-func" primitive rather than a 5th copy next time.

Leaving 1–3 as follow-up rather than expanding this PR further; #2 is the one I'd prioritize if picked up.

Issue #333, opened 2026-08-11
guard: gated pre-write lint substrate — run ruff against a proposed Write payload before it reaches disk


inth3shadows:
## Observation

Every hook-based linter integration surveyed lints **after** the write —
Pyright/Claude-Code hooks, pi-hooks/lsp, zircote's plugin, Cursor, Windsurf. The
model has already been told the edit succeeded before anything checks it. RunEcho
is the only surface in this repo that sees proposed content *before* it exists,
and it currently spends that position on symbol resolution alone.

A trusted whole-file linter would extend the pre-write moment to a much larger
finding class at essentially zero implementation cost per rule — but it does not
fit the hook the way the existing checks do, for two specific reasons, and both
have to be designed around rather than discovered later.

## Constraint 1: the guard sees a hunk, ruff needs a file

`cmd/runecho-guard/main.go:724` states it: *"An Edit/MultiEdit hunk sees only the
changed region, not the rest of the file."* A whole-file linter needs the full
proposed content, which for Edit/MultiEdit means reading the pre-edit file off
disk and splicing `new_string` in.

That reconstruction races. This repo's own working pattern is multiple Claude
Code sessions pushing to the same tree in parallel, and PreToolUse fires *before*
the write — so the disk state read at splice time may already be someone else's.
The result is a lint finding about a file state that **never existed on disk in
that form**.

The second-order consequence is the disqualifying one. `cmd/runecho-ir/fpaudit.go`
judges every finding against git history (`git rev-list` / `git rev-parse` /
`git grep`; see its header, and `internal/guardstats/gitoracle.go:257`). A
phantom state never enters history, so those findings rate permanently
`VerdictUnknown` — or, worse, read as never-defined and score `VerdictStands`,
which `internal/guardstats/gitoracle.go:26` notes *"flatters the guard."* The
telemetry that would tell us whether the substrate works is the telemetry the
naive implementation poisons.

**So v1 is Write-only.** A Write payload's `content` field IS the complete
proposed file — the repo already relies on this at `main.go:1010` (*"Not needed
for Write: its newLines already IS the whole file"*). No disk read, no splice,
no race. Verified: ruff takes the content on stdin, so nothing is written
anywhere:

```
printf 'import os\n\ndef f():\n    return undefined_thing(os.sep)\n' \
  | ruff check --no-cache --isolated --select F821,F811 \
      --stdin-filename /path/to/proposed.py --output-format json -
[...]

Issue #323, opened 2026-08-07
spike: optional external-oracle composition for guard checks (lsp-go / Ctxo) -- measure hit-rate vs latency


inth3shadows:
## Context

Brainstorm option A (2026-08-07, scored F=4/N=3): instead of racing to deepen RunEcho's own type-resolution engine (the "strategic danger" path #312 itself warns about -- local types -> parameters -> return types -> cross-file -> interfaces -> generics -> call graph, i.e. reimplementing go/types), keep the fast standalone IR as the default and optionally call an external oracle for a stronger verdict when one is available and cheap enough, falling back cleanly when not.

## Research (searxng + local LiteLLM subagent, 2026-08-07, gh-verified)

No prior art found for exactly this tiered fast-gate-plus-optional-oracle architecture -- existing LSP-via-MCP work (three separate writeups found) is all read-side navigation, not a blocking gate with fallback. One good, verified concrete oracle candidate did turn up: alperhankendi/Ctxo (67 stars, pushed 2026-08-02, npm-installable MCP server). Confirmed via its own README: it's a read-side deterministic graph (blast radius, git intent, 14 semantic MCP tools) -- NOT a competing pre-write gate, so it's safe to treat as a potential backend rather than a rival. You also already have a working same-shape bridge installed: the lsp-go MCP (mcp-language-server wrapping gopls, verified end-to-end 2026-07-18).

## Spike

When RUNECHO_GUARD_VARTYPE's local Go var-type inference fails to resolve a v.Method() call (#312 slice 1 currently at 15/60 on the x/text corpus), optionally call the lsp-go bridge (or Ctxo) for a stronger verdict behind a feature flag, instead of giving up. Measure against the same corpus:

- Does it move the hit rate further than 15/60?
- What's the added p50 latency against the guard's ~12ms hook budget (named directly in #312 as the real constraint)?

**Pivot point:** if the call adds more than ~5-8ms p50, synchronous composition doesn't fit the budget -- park this and treat it as async/best-effort-only, or abandon in favor of the positioning (see the sibling issue on fpaudit-led positioning) and contracts (#12) lanes instead.

## Non-goals

Not a commitment to ship a permanent dependency on an external oracle. This is a measurement spike to test whether "compose with an oracle" beats "build our own resolver" on the exact metric #312 already tracks.

inth3shadows:
## Spike result — measured, not just prototyped

[...]

Issue #313, opened 2026-08-06
Extend the compiler-oracle differential to Python (ruff F821)


inth3shadows:
#308 built the differential for Go, with `go build` as adjudicator. Python has a usable CLI oracle already installed: `ruff check --select F821 --output-format json` reports undefined names in ~19ms/file.

Same two phases, same posture stratification:

- **False positive** — a file ruff passes clean proves the names resolve, so a guard flag there is provable.
- **False negative** — rewrite a use site to a fresh name, confirm ruff now reports F821 at that line, then check whether the guard says anything.

Python is where the measured false-positive rate is highest (~5% on the corpus, vs ~0% for Go), and where the corpus is most likely to share its author's assumptions — so it is the language where an external oracle should tell us the most.

Two things to carry over from the Go build, both learned the hard way:

1. **Measure every posture.** The Go harness first passed the whole file as both the added text and the fold source, so the known set always already contained every declaration in the file. It reported zero false positives while two default-on paths were broken. Postures needed: whole-file Write, Write-creating-a-file (no fold), Edit hunk, pre-commit staged diff.
2. **Stratify by reference shape.** A single blended rate hides which shape is blind, and the shape is the fix.

Caveat specific to ruff: F821 abstains on star-imports and some dynamic forms, so it is a *conservative* oracle — silence is weaker evidence than a clean `go build`. Classify "oracle silent" separately from "oracle confirms resolved" rather than counting the former as a proven false positive.

TypeScript (`tsc --noEmit`, TS2304) is the same shape but needs a per-repo `tsconfig.json`; tracked separately if this lands well.

PR #340, opened 2026-08-12
guard: oracle-differential harness for the lint substrate (#333 P2-2)

inth3shadows:
> **Numbers corrected in the third commit (`33c159e`).** A code-review pass found that zero-byte `.py` files bypassed the hook and still counted as observations, and that a nearest-rank p99 on a small sample is just the maximum. The table below is the corrected one; the p99 column is withdrawn as unstable rather than restated. See the commit message and https://github.com/inth3shadows/runecho/issues/333#issuecomment-5261779801 for the full accounting.

Closes the validation half of #333. Test-only — no shipped-code change, deliberately: the `RUNECHO_GUARD_LINT=1` dogfood window opened 2026-08-11 and a behavior change would reset it.

## What it does

Validates the gated pre-write lint check by measurement rather than user recruitment: posture fidelity, overlap partition against the existing checks, and latency, all computed in the harness.

It drives the real `runHookMode` and adjudicates against `ruff` invoked independently — on the file as a **path argument** while the check uses `--stdin-filename`. If both sides went through stdin a stdin-plumbing bug would cancel out on both sides of the comparison and read as agreement.

```
RUNECHO_LINT_CORPUS=<repo> RUNECHO_LINT_CORPUS_MAX=250 \
  go test ./cmd/runecho-guard -run TestLintDifferential -v
```

## The posture correction

Three postures, identical payload, differing only in what sits on disk — the thing ruff never sees but `readFileLines` folds into the additive check.

| Posture | On disk | Purpose |
|---|---|---|
| `create` | absent | additive check sees only the payload |
| `overwrite-same` | byte-identical | **control**, asserted equal to `create` |
| `stale-fold` | defines the names the payload no longer does | a rewrite that drops a definition it still calls |

The first cut used only the first two, on the theory that a byte-identical on-disk file would quiet the additive check. It does not — folding a file's own defs adds nothing when the payload already contains them — so the two produced identical partitions. One posture measured twice, wearing the label of stratification: #313's trap, reproduced by the harness written to defend against it.

`stale-fold` is the one that separates them. On the committed corpus it moves the partition from `both=2 ruff-only=2` to `both=0 ruff-only=4`.

## Mutation-tested, not just run

Two deliberate breaks both fail it:
- linting the on-disk file instead of the payload (#313's exact trap) — caught by `stale-fold` for the F821 cases and by **nothing else**
- a +1 line offset on the reported finding

## Results (re-measured 2026-08-12)

| Corpus | Observations | Oracle findings | Hook FPs | Marginal p50 |
|---|---|---|---|---|
| `testdata/lintcorpus` | 15 | 12 | 0 | n too small to publish |
| `secret-broker` | 750 | 0 | 0 | +9.2 ms |
| `competitive-intel` | 750 | 0 | 0 | +11.0 ms |

p50 only: re-running the unchanged harness gave p99 of +30.5, +13.6 and +13.1 ms on the same corpus, so the tail measures the workstation as much as the check.

The committed corpus no longer publishes a marginal p50 at all. It previously read `+9.9 ms`, but four runs of the *unchanged* harness spread it across +7.3, +12.9, +8.8 and +13.9 ms — n=15 is too small for the median, not just the tail. That corpus exists to prove a true finding survives the plumbing end-to-end, which the F821-clean real corpora cannot do; it is not a latency measurement.

The two runs are now **order-alternated** per file so neither side systematically pays a payload's cold-cache costs. Under the original fixed ON-then-OFF order the same corpora read +10.9 and +11.0 ms: alternating moved `secret-broker` by 1.8 ms and `competitive-intel` not at all, which puts the ordering inside run-to-run noise rather than making it the upward bias it looked like.

**Latency:** the check roughly doubles the hook's p50 and is worse at the tail. That is the measurement behind shipping default-off, now taken in the harness rather than by hand, and it fails #323's ~5-8 ms pivot criterion decisively.

**Committed Python is F821/F811-clean.** 1,500 observations across two real repositories, not one oracle finding — an undefined name is a bug someone fixed before committing. So no corpus can validate this check's positive direction. The harness distinguishes the two cases rather than reporting one number: a silent oracle is a hard failure on the committed corpus and an explicitly-labelled "false-positive freedom and latency only" run on any other. `0 false positives over 750 observations` is the exact sentence a completely disabled check also produces.

## Placement note

Lives in `package main`, not `internal/guard` where the issue names it. `lint.go` ships in `cmd/runecho-guard`, so `internal/guard` cannot reach `lintFindingsWithReason`, `suppressAlreadyReported` or `runHookMode` — a harness there could only exercise a copy of the ruff call written for the test, which is the copy-not-the-shipped-path failure above. Revisit if #313 moves the lint core.

## Second commit

Three defects found by self-review after the first commit: a latent stale decision-log read (the log is append-only and every observation targets the same path, so a run that skipped `logDecision` would leave the previous file's symbols as the last line, with a matching `file` field — undetectable misattribution), a dead map, and a hand-rolled `exec.ExitError` assertion replaced with `errors.As`. Re-measurement confirms the numbers are unchanged.

## Third and fourth commits — two review rounds

Round 1 found seven defects, fixed in `33c159e`. A second review pass then found nine more, all closed in `eb7439b`.

The one that mattered: **the harness's header claimed it asserted suppression, and it did not.** Both fidelity directions treated `suppressAlreadyReported` as an excuse to *skip* an unreported oracle finding, which asserts nothing — if suppression did nothing, the finding would simply arrive reported and match. Mutation-checked: no-op'ing the `suppressAlreadyReported` call left the entire harness green while a real hallucination was printed under two headers and double-counted in `decisionRecord.Symbols`. That is the same "presence is not verification" class this harness was written to defend against, reproduced one level up. The invariant is now stated directly — a *reported* finding whose symbol another check also flagged **is** that double-count — and the mutant dies with four named failures.

The rest: both directions are multiset-keyed rather than set-keyed (a set let the same finding render twice and match an already-matched key; mutation-checked with a double-render mutant); `oracleAbstains` strips trailing comments, so `from x import *  # noqa` no longer misfiles ruff's module-wide silence as a real divergence, pinned by a new table-driven test; an unadjudicable file drops itself instead of aborting a 750-observation sweep; a typo'd `RUNECHO_LINT_CORPUS` reports the walk error instead of "contains no .py files"; the partition log states occurrence and distinct cardinalities on one line and carries file attribution; and the fidelity summary no longer prints `OK` underneath its own failures.

Two documentation contradictions were resolved by re-measuring rather than by picking a side. The empty-`__init__.py` counts disagreed between the test comment and TECHNICAL.md — re-counted with the walker's own exclusions at **18 of 381** and **21 of 453**; the larger figure had come from a bare `find` that counted the vendored trees `lintCorpusSkipDirs` skips. And the reproduction command passed `MAX=300` above a 250-file table.

## Verification

- `go build ./...` and `go test ./...` green
- gofmt clean under CI's Go 1.25 toolchain
- `python3 bench/hookmutate/mutate.py --strict` exit 0, 8 survivors all pre-existing documented `known_gap`s


inth3shadows:
## Round-2 review findings — recorded, NOT fixed

Round 1's seven findings are fixed in `33c159e`. A second review pass then found nine more. `/autorun` caps me at two review rounds, so these are documented rather than fixed — the PR is held for manual merge anyway (`release.yml` fires on push to `master`), so nothing lands on these unaddressed.

Ordered by what I'd fix first.

**1 — medium. The harness does not actually assert suppression, though its header says it does.** Both fidelity directions treat `suppressAlreadyReported` as an *excuse to skip* (`if _, dup := o.other[f.Symbol]; dup { continue }`), never as an assertion. Mutation-verified by the reviewer: replacing the `suppressAlreadyReported` call in `main.go:968` with a no-op leaves `TestLintDifferential` **fully green**, while `undefined_name.py` emits `missing_helper` under two headers and double-counts it in `decisionRecord.Symbols`. Only `lint_test.go`'s 2-line synthetic payload kills that mutant. Missing assertion: a *reported* finding whose Symbol is in `o.other` is a double-count and should fail. This is the same "presence is not verification" class the harness was written to defend against, so it is the one I'd want closed before merge.

**2 — medium. `oracleAbstains` is defeated by a trailing comment.** It requires the line to *end* with `" import *"`, so `from os.path import *  # noqa: F403` reads as non-abstaining. Reproduced: ruff still abstains module-wide, but those observations land in `guard-only/oracle-clean` — the bucket TECHNICAL.md calls a real divergence to investigate. The committed `star_import.py` classifies correctly only because its star import happens to end the line.

**3 — low.** `byKey`/`reportedKeys` are sets, not multisets, and nothing bounds `totalReported` against `totalOracle`, so rendering the same finding twice passes both directions. Same double-counting class as #1 from the other side.

**4 — low.** The lint-ON run always precedes lint-OFF for a given payload, so that observation's cold-cache costs are paid by the ON run and credited to the check — biasing the published marginal cost upward. Worth noting: round 1 tested this hypothesis directly and *falsified* it with an instrumented run (cold-position OFF p50 11.4 ms vs warm-position 11.3 ms). The cheap resolution is to alternate the order and stop arguing about it. Direction is conservative either way — it overstates the cost of a check that already ships default-off.

**5 — low.** `ruffOracle` calls `t.Fatalf` on any ruff exit other than 0/1, so one permission-denied `.py` aborts a 750-observation sweep. Contradicts `lintCorpusFiles`, which deliberately skips unreadable subtrees.

**6 — low.** `filepath.WalkDir`'s root error is swallowed, so a typo'd `RUNECHO_LINT_CORPUS` fails with "contains no .py files — the harness would pass vacuously" instead of "no such directory", pointing the reader at the corpus when the fault is the env var.

**7 — low.** `guardOnlyClean` is an occurrence count while `cleanSample` is a de-duplicated symbol set; they are logged adjacently as if the names explain the count. The names also carry no file attribution, so the bucket the docs call "the next investigation" cannot be triaged from the log.

**8 — low.** Contradictory empty-`__init__.py` counts: the test comment says "18 of 378 in one real corpus, 63 of 3049 in another"; TECHNICAL.md says "18 and 21". The harness logs tree-wide (pre-cap), so both cannot be right. My 63 came from a `find` that used different exclusions than the walker.

**9 — low.** TECHNICAL.md's reproduction command passes `RUNECHO_LINT_CORPUS_MAX=300` but the table above it is 250 files, so running the documented command produces different numbers than the table it sits under.

### Cleared on inspection

`percentile`'s nearest-rank arithmetic and the `tailLabel` guard; the `lintDiffEnv` env audit (`RUNECHO_GUARD_QUALIFIED` genuinely is the only inverted gate); the decision-log truncation fix in `0452ade`; `assertPostureControl`'s equality, which held across 250 diverse third-party files; and CI's `ruff==0.16.1` install, so `requireRuff`'s skip does not silently drop coverage there.

### Coverage note, not a finding

The harness always enrolls a store, so `hookdegraded.go`'s independent lint plumbing (different content expression, `askWithoutIndex` renderer, no suppression) is never reached by the differential. `lint_test.go` covers it as a unit test, so it is not uncovered — just outside what this harness measures, despite TECHNICAL.md describing unenrolled trees as the common case for a globally installed hook.

inth3shadows:
## Round-2 findings — all nine closed in `eb7439b`

Fixed in the order they were ranked.

**1 — the missing suppression assertion.** Stated directly now: a *reported* lint finding whose symbol another check also flagged for the same edit is the double-count `suppressAlreadyReported` exists to prevent. Mutation-verified both ways — the reviewer's no-op mutant on `main.go:968` previously left the harness fully green; it now dies with four named failures (`unresolved_parser`, `missing_helper`, each under two postures). Guarded on `Symbol != ""` because `lint.go` documents itself as never suppressing an unparsed message.

**2 — `oracleAbstains` defeated by a trailing comment.** Strips the comment (and a trailing `;`) before matching. Pinned by `TestOracleAbstains_TrailingComment`, table-driven rather than a new corpus fixture — a fixture would have to be ruff-adjudicated and would move the published observation counts to test a log label. Over-abstaining on a `from …` line inside a triple-quoted string is possible and is the safe direction, noted at the call site.

**3 — sets, not multisets.** Both directions now compare occurrence counts per `line:rule:symbol` key, which is strictly stronger than the global `totalReported <= totalOracle` bound the finding asked for and localises the failure. Mutation-verified with a double-render mutant.

**4 — fixed ON-then-OFF ordering.** Alternated per file index. Worth recording: this **did** move a number — `secret-broker` went +10.9 → +9.2 ms — but `competitive-intel` did not move at all (+11.0 both ways), so the ordering sits inside run-to-run noise rather than being the upward bias it looked like. Alternated regardless; it costs nothing and closes the argument.

**5 — `t.Fatalf` on an unexpected ruff exit.** An unadjudicable file now drops itself and is logged; the sweep continues. A systemic break still fails loudly — `len(adjudicated) == 0` is fatal.

**6 — swallowed walk root error.** `path == root` returns the error, so a typo'd `RUNECHO_LINT_CORPUS` reports the walk failure instead of "contains no .py files".

**7 — occurrence count logged beside a distinct set.** Both cardinalities now print on one line, and `cleanSample` is keyed `symbol@file` so the bucket the docs call the next investigation can actually be triaged.

**8 — contradictory empty-`__init__.py` counts.** Re-counted with the walker's own exclusions: **18 of 381** (`secret-broker`) and **21 of 453** (`competitive-intel`). Your 63/3049 came from a `find` that counted the vendored trees `lintCorpusSkipDirs` skips. TECHNICAL.md's "18 and 21" was right; the test comment was wrong.

**9 — repro command vs table.** `MAX=250`, matching the table above it.

### Beyond the nine

The fidelity summary logged `posture fidelity OK` underneath its own `t.Errorf` output — visible in the mutation run. Guarded on `t.Failed()`.

Re-measuring for #4 also showed the committed corpus is too small to publish *any* latency figure from: four runs of the unchanged harness spread its marginal p50 across +7.3, +12.9, +8.8, +13.9 ms. The table published +9.9 ms as if that were stable. It now publishes nothing there and says why — the corpus exists to prove a true finding survives the plumbing, which the F821-clean real corpora cannot do.

Full suite green; `gofmt` clean under the CI-pinned Go 1.25 toolchain, not just local 1.26.

PR #341, opened 2026-08-12
guard: pin findings 5-7 of the lint differential by mutation (#333)

inth3shadows:
Follow-up to #340. Three of that PR's nine review findings — #5, #6, #7 — were closed by inspection and pinned by nothing. This holds them down by mutation, and in doing so corrects finding #5's premise.

Test-only, same as #340: no shipped-code change, so the `RUNECHO_GUARD_LINT=1` dogfood window opened 2026-08-11 is not reset.

## The gap was measured, not assumed

Reverting each of the three fixes on `master` left the whole package green:

| Mutant | Before this PR |
|---|---|
| `ruffOracle` aborts the sweep again | **survived** |
| walk root error swallowed again | **survived** |
| `cleanSample` back to a bare symbol key | **survived** |

## Finding #5's stated failure cannot happen — the real one is quieter

The review said a permission-denied `.py` would `t.Fatalf` and take a 750-observation sweep down with it. ruff does not report an unreadable file through its exit code. Measured against 0.16.1, a `chmod 000` file that **genuinely contains an F821**:

```
exit 0    stdout: []    stderr: warning: Failed to lint <path>: Permission denied (os error 13)
```

Read by exit code alone that is indistinguishable from a clean file. So it entered the corpus as an adjudicated-and-clean observation the oracle never read a byte of — a silent entry in the denominator of every "0 false positives over 750 observations" claim. The same false-denominator defect as the zero-byte files `lintCorpusFiles` already drops, one layer down.

This was found by trying to write the test, not by reading the code.

The oracle now captures stderr and treats a read failure — and **only** a read failure — as unadjudicable. Dropping on any stderr at all would silently shrink the corpus every time ruff emitted a deprecation or unused-noqa warning, which is a loss of coverage that reads as a smaller corpus rather than as a fault. Both directions are pinned.

## Seven mutants, all killed

Three of them are **over-fix** mutants: they pin the boundary, not just the behaviour.

| # | Mutant | |
|---|---|---|
| F5a | drop the stderr read-failure check | killed |
| F5b | abort the sweep on an unexpected exit | killed |
| F5c | *over-fix:* any stderr means unadjudicable | killed |
| F6a | swallow the walk root error | killed |
| F6b | *over-fix:* surface every walk error, not just the root's | killed |
| F7a | `cleanSampleKey` drops the file | killed |
| F7b | `cleanSampleKey` drops the symbol | killed |

F5c and F6b carry real weight. The first would quietly shrink the corpus on a benign warning; the second would make any repository containing one unreadable directory unmeasurable. Both are the failure mode where "fixed" silently costs coverage.

## Three seams, each justified by unreachability

- **`walkLintCorpus`** split out of `lintCorpusFiles`, because `lintCorpusFiles` reports every fault through `t.Fatalf` and a test cannot tell one fatal from another — which is exactly how the swallowed root error hid, surfacing as the vacuous-corpus message about a different fault entirely.
- **`cleanSampleKey`** extracted, so the oracle-clean bucket's file attribution is asserted data rather than log format.
- **`ruffBin`** made a var, because real ruff cannot reach the unexpected-exit branch — the harness always passes an absolute path, so no filename can be misread as a flag. A branch whose comment claims "one oddity cannot take the sweep down" while nothing can verify it is the thing this harness exists to object to.

## What is deliberately not pinned

The other half of finding #7 — the occurrence-vs-distinct **log labels** — is `t.Logf` text. Pinning it means capturing and asserting on log output, a contrived harness for a label. The data property (the key carries both symbol and file, so the same name in two files does not collapse) is pinned; the formatting is not, and this says so rather than implying the finding is fully held down.

## Review round

Four findings, all in the new test code, all fixed in `adeba51`.

The sharpest one landed on this PR's own thesis: **`TestRuffOracle_UnexpectedExitDropsOneFileNotTheSweep` could pass while asserting nothing.** `ok == false` is also what an *unexecutable* stub produces — a noexec `TMPDIR` or a missing `/bin/sh` makes `cmd.Run` return an `*exec.Error`, which fails the `*exec.ExitError` assertion and reaches the same `return nil, false` without the exit code ever being the reason. A vacuous negative, written one function after the sibling test that deliberately guards against exactly that with a readable control. It now runs an **exit-0** stub first and skips if that cannot execute, so the exit-2 negative is known to be caused by the exit code; both variants emit identical valid JSON, so the exit status is the only difference between them.

Both chmod-based skips gated on `os.Geteuid() == 0`, which is not the only way `chmod 000` fails to deny access: `CAP_DAC_OVERRIDE` grants it to a non-zero uid under containers and rootless/userns, and mode bits are advisory on a drvfs/NTFS `TMPDIR` under WSL. In those environments both tests reported spurious failures rather than skipping. They now **probe** — `os.ReadFile` after the file chmod, `os.ReadDir` after the directory chmod — and skip on the result instead of inferring it from the uid.

Moving `ruffBin` in front of `ruffOracle` had fused `ruffOracle`'s entire doc comment onto a one-line `var`, leaving the function undocumented. The var and its comment now precede the doc block rather than splitting it.

The matrix was re-run after these changes rather than assumed to hold — adding skip paths is precisely how a kill silently becomes a survivor. Still 7/7, and all five tests report PASS rather than SKIP in the development environment, so the new probes are not disarming them there.

## Verification

- `go build ./...` and `go test ./...` green
- gofmt clean under CI's Go 1.25 toolchain, not just local 1.26
- `python3 bench/hookmutate/mutate.py --strict` exit 0, 8 survivors all pre-existing documented `known_gap`s — unchanged from #340
- mutation matrix re-run post-review: 7/7 killed, no kill weakened by the new skip paths
- 250-file real-corpus run (`secret-broker`): zero files dropped, 750 observations, marginal p50 +10.6 ms — inside the ~2 ms run-to-run spread #340 documented


inth3shadows:
## Review findings — all four fixed in `adeba51`

**1 — the exit-code test could pass while asserting nothing.** Correct, and the sharpest of the four: `ok == false` is also what an unexecutable stub produces. A noexec `TMPDIR` or a missing `/bin/sh` makes `cmd.Run` return an `*exec.Error`, which fails the `*exec.ExitError` assertion and reaches the same `return nil, false` without the exit code ever being the reason — a vacuous negative in a PR whose entire thesis is that vacuous assertions are the defect. It now writes an **exit-0** stub first and skips if that cannot execute, so the exit-2 negative is known to be caused by the exit code. Both variants emit identical valid JSON, so the exit code is the only difference between them.

Worth noting the escape hatch that already existed but should not be relied on: in a noexec environment `TestRuffOracle_BenignStderrStillAdjudicates` would have failed loudly, so the package would not have gone green in silence. That is a neighbouring test catching it, not this one asserting it, which is not the same thing.

**2 and 3 — `os.Geteuid() == 0` is the wrong predicate.** Also correct, and it matters on this box specifically: mode bits are advisory on a drvfs/NTFS `TMPDIR` under WSL, and `CAP_DAC_OVERRIDE` grants reads to a non-zero uid under containers and rootless/userns. Both tests now **probe** — `os.ReadFile` after the chmod, `os.ReadDir` after the directory chmod — and skip on the result rather than inferring it from the uid. Strictly better than the uid check in every environment, so the uid check is gone rather than kept alongside.

**4 — the fused doc comment.** Real regression I introduced: moving `ruffBin` in front of `ruffOracle` attached the whole of `ruffOracle`'s rationale to a one-line `var` and left `ruffOracle` undocumented. The var and its comment now precede the doc block rather than splitting it.

### On the note about `"Failed to lint"` being ruff-gated

Agreed, and left as is. `TestRuffOracle_UnreadableFileIsNotAdjudicated` skips without ruff, so F5a would survive a local run on a machine without it — but CI installs `ruff==0.16.1` on non-root `ubuntu-latest`, so the pin is live where it gates merges. Same posture as every other ruff-dependent assertion in the file; making this one the exception would be inconsistent rather than safer.

### Re-verification

Mutation matrix re-run after the changes — the skips introduced above are precisely the kind of thing that can silently convert a kill into a survivor:

| # | Mutant | |
|---|---|---|
| F5a | drop the stderr read-failure check | killed |
| F5b | abort the sweep on an unexpected exit | killed |
| F5c | *over-fix:* any stderr means unadjudicable | killed |
| F6a | swallow the walk root error | killed |
| F6b | *over-fix:* surface every walk error | killed |
| F7a | `cleanSampleKey` drops the file | killed |
| F7b | `cleanSampleKey` drops the symbol | killed |

Still 7/7. All five tests report PASS rather than SKIP in this environment, so the new probes are not silently disarming them here. Full suite green; gofmt clean under the CI-pinned Go 1.25 toolchain.

PR #344, opened 2026-08-12
DO NOT MERGE: prove the skip allowlist gate fails red

inth3shadows:
# DO NOT MERGE — this branch is a probe

Proof-of-red for the skip allowlist gate merged in #343. Close this PR once CI has gone red; it must never land.

## Why it exists

#343's two failure directions were verified locally against staged logs, and its CI run exercised only the happy path — the runner's skip set matched the allowlist, so the gate passed. "It fails when it should" was therefore still an untested claim about the gate itself.

That is the exact pattern this repo keeps finding inside its own instruments: an assertion-shaped thing that has never been observed asserting. #333's harness looked green while `suppressAlreadyReported` was a no-op; #341's exit-code test passed while its stub could have failed to execute. A gate that has never been seen red belongs in the same category until it is.

## What it changes

One file, no code touched, both directions at once:

| direction | change | expected verdict |
|---|---|---|
| **unexpected skip** | removed the `TestFileScopeFPSmoke` entry | its genuine skip is no longer allowlisted → FAIL |
| **stale entry** | added `TestThisDoesNotExistAnywhere` | cannot skip, because it does not exist → FAIL |

Local dry-run against the real merged CI log from #342 produces exactly that, exit 1:

```
**FAIL — skipped but not in the allowlist.** ...
TestFileScopeFPSmoke

**FAIL — listed as expected-to-skip but did not skip.** ...
TestThisDoesNotExistAnywhere
```

The interesting part is what the runner does, not what my laptop does.

## Why this is a PR and not a master push

Deliberate, and the reasoning is worth recording because the obvious version of this experiment is genuinely destructive here.

`autorelease` (`release.yml:55`) has **no `needs:`** on the test workflow. It does not wait for CI, does not consult it, and fires on any push to master. `bump_for` (`next-release.py:74`) returns `patch` for any subject without a `docs`/`chore` prefix. So proving this on master would have tagged and **published a real release** — binaries, Sigstore attestations, CHANGELOG roll — from a commit broken on purpose, and the revert would have published a second one.

A pull request cannot reach `autorelease` at all, and `release.yml`'s `pull_request` trigger is path-filtered to the goreleaser files, so that workflow does not even run for this branch. Same evidence, zero blast radius.

## Expected outcome

`test` job **red**, failing at "Gate skipped tests against the allowlist", with both failure blocks in the run summary. `docker` should stay green — it shares no steps with this.

If the gate passes here, the gate is broken and #343 needs reverting.


inth3shadows:
## Result: the gate fires. And the probe found a defect.

`test` **failed in 6m32s**, at the right step:

| step | outcome |
|---|---|
| Test (race + coverage, includes guard corpus gate) | success |
| **Gate skipped tests against the allowlist** | **failure** |
| Installer syntax | not reached |
| Hook-fixture mutation gate | not reached |

The suite itself passed, so the build went red *because of the gate* rather than collateral test damage. `docker` stayed green as expected. Both failure directions were staged in one commit, and the gate's exit code condemned the run.

### The defect: a red gate says nothing in the job log

```
##[group]Run bash .github/scripts/check-skips.sh test.log .github/expected-skips.txt
shell: /usr/bin/bash -e {0}
##[endgroup]
##[error]Process completed with exit code 1.
```

That is the entire step output. Which test skipped unexpectedly, which entry was stale — all of it went to `$GITHUB_STEP_SUMMARY`, which is a **file**, not stdout. A maintainer looking at a failed run in the log sees an opaque exit code and has to already know to open the summary page.

This is invisible under local testing, which is the interesting part: `GITHUB_STEP_SUMMARY` is unset on a developer box, so `emit()` falls back to `/dev/stdout` and every local case printed its reasons perfectly. The nine local cases all passed *and* all missed this. The behaviour only diverges in the one environment that matters.

Which is precisely the argument for running this probe at all. #343's gate was verified by nine local cases and a green CI run, and was still shipping a failure mode nobody would have seen until the first genuine red build — probably months from now, on a run where the cause was not already known.

Fix: build the report once and write it to both the summary and stdout, so the log carries the reason too. Then re-run this probe against the fixed script — a fix to failure output that has not been observed failing is the same untested claim one layer down.

Closing unmerged, as intended. Branch kept for the moment so it can be rebased onto the fix and re-run.

PR #346, opened 2026-08-12
DO NOT MERGE: re-prove the skip gate fails red, with log output

inth3shadows:
# DO NOT MERGE — this branch is a probe

Second run of the #344 probe, now against the log-output fix merged in #345. Close once CI has gone red; it must never land.

## Why a second run

#344 proved the gate fails red, and in doing so found that a failing gate printed **nothing** to the job log — only `Process completed with exit code 1`. Every diagnostic went to `$GITHUB_STEP_SUMMARY`, a file, so the reason was reachable only by knowing to open the summary page.

#345 fixed that: the report is built once and flushed to both the summary and stdout. Its own CI run confirms the **success** path now reaches the log:

```
### Skipped tests

3 expected skip(s), all allowlisted:
TestFileScopeFPSmoke / TestSpikeGoplsOracleVarType / TestSpikePrematureLatency
```

But the defect was in the **failure** path, and #345's green run cannot exercise it. A fix to failure output that has never been observed failing is the same untested claim one layer down — which is exactly what #344 caught the first time. So the probe runs again.

## What it changes

Identical to #344, rebased onto the fix. One file, no code touched:

| direction | change | expected verdict |
|---|---|---|
| **unexpected skip** | removed the `TestFileScopeFPSmoke` entry | its genuine skip is no longer allowlisted → FAIL |
| **stale entry** | added `TestThisDoesNotExistAnywhere` | cannot skip, because it does not exist → FAIL |

A fresh branch rather than a force-push of #344's, so the original probe and its findings stay intact in history.

## Expected outcome

`test` **red** at "Gate skipped tests against the allowlist" — and this time the job log itself must contain both failure blocks naming `TestFileScopeFPSmoke` and `TestThisDoesNotExistAnywhere`. If the step is red but the log is still bare, #345 did not work and the fix needs revisiting.

Local dry-run in the CI shape (`GITHUB_STEP_SUMMARY` set to a real file) produces both blocks on stdout and exits 1. The runner is the part that counts.

Same reasoning as #344 for why this is a PR and never a master push: `autorelease` (`release.yml:55`) has no `needs:` on the test workflow and fires on any push to master, and `bump_for` (`next-release.py:74`) defaults to patch — so this on master would publish a real release from a knowingly-broken commit.


inth3shadows:
## Confirmed: the fix works on the path that mattered

`test` failed in 6m24s at "Gate skipped tests against the allowlist", and this time the **job log itself** carries the reason:

```
### Skipped tests

**FAIL — skipped but not in the allowlist.** ...
TestFileScopeFPSmoke

**FAIL — listed as expected-to-skip but did not skip.** ...
TestThisDoesNotExistAnywhere

##[error]Process completed with exit code 1.
```

Compare #344 on the same probe before #345, where the whole step output was:

```
##[error]Process completed with exit code 1.
```

Both failure directions fired, both are named, and a maintainer hitting this red build learns the cause without knowing to open the run summary page.

### Why the second probe was not redundant

#345's own CI run was green, and it demonstrated the fix only on the SUCCESS path — the log showed `3 expected skip(s), all allowlisted`. But the defect was entirely in the FAILURE path, which a green run cannot reach. Shipping on that evidence would have been the same shape of mistake #344 caught the first time: a claim about failure output, verified only by watching it succeed.

Closing unmerged, as intended. `ci-gate-proof-do-not-merge` (#344) and this branch both stay closed and unlanded; the probe is cheap enough to recreate whenever the gate changes.

Issue #351, opened 2026-08-14
history.db grows unbounded: hourly snapshots with no change-dedup or retention


inth3shadows:
## Summary

`~/.runecho/history.db` grew to 4.5GB on a work Mac (84 enrolled repos, first enrollment 2026-06-07), which caused every `runecho-mcp` invocation — including `--version` — to hang for ~30s before dying silently with no output. That silent 30s death is also what made a downstream `terse` MCP connection time out.

A second machine (WSL, this box) shows the same trajectory: `history.db` already at 1.4GB with 526 enrolled repos, 5,078 snapshots.

## Root cause (verified against the Mac's DB directly)

A background refresh job re-indexes every enrolled repo on an **hourly cadence**, forever, since first enrollment. Confirmed via `snapshots.timestamp` for one repo: `23:04:23 → 00:05:59 → 01:06:00 → 02:06:01 → 03:06:02 ...` — a new row every ~61 minutes, with no gaps, continuing to the present.

The bug: **each hourly tick writes a brand-new `snapshots` row (plus fresh copies of `files`/`symbols`/`refs`) even when nothing changed.** For the repo checked:

- 1,651 total snapshots
- only **138 distinct `root_hash` values** — i.e. the repo's actual content changed 138 times, not 1,651
- the most common `root_hash` alone repeats **382 times** unchanged

So ~92% of snapshots for that repo are pure duplicates of content already stored. There's also no retention/pruning — nothing ages out, ever.

Extrapolated across 84 repos over 68 days: 83,721 snapshots, 2.08M `files` rows, 28.3M `symbols` rows, 14.2M `refs` rows → 4.5GB.

## Suggested fix

1. **Skip the write** when the new snapshot's `root_hash` matches the most recent snapshot for that `repo_id` — update `last_indexed` on the existing row instead of inserting a new one.
2. **Retention policy** on `snapshots` (and cascade to `files`/`symbols`/`refs`) — e.g. keep the last N per repo, or age out anything older than N days, whichever a maintenance pass enforces periodically.
3. Separately (robustness, not root cause): `runecho-mcp` should not die **silently** after ~30s when it can't open/load `history.db` in time — at minimum log to stderr so this doesn't look identical to a hung process from the outside.

## Repro / verification queries

```sql
-- snapshot count vs distinct content for one repo
select count(*), count(distinct root_hash) from snapshots where repo_id = <id>;

-- duplicate rate
[...]

PR #352, opened 2026-08-15
fix(mcp): answer --version before opening the store (#351)

inth3shadows:
Part 1 of 4 for #351. This is the robustness item (#351 fix 3) — smallest blast radius, zero interaction with the snapshot store, so it ships first.

## The bug

`cmd/runecho-mcp/main.go` never read `os.Args`. There is no flag parsing in this binary at all, so `runecho-mcp --version` fell through to `run()`, opened `history.db` at line 41, and then blocked in `server.Serve` at line 53 waiting for a JSON-RPC `initialize` frame that a version caller never sends.

From the outside that is indistinguishable from a hang. It is also the observed symptom in #351: every `runecho-mcp` invocation *including `--version`* appearing to hang for ~30s before dying silently, which in turn timed out a downstream `terse` MCP connection.

Two independent causes were compounding:

- **Unconditional** — a version probe had no business opening a database. Fixed here.
- **Size-dependent** — `snapshot.Open` runs `integrityCheck()` → `PRAGMA quick_check`, which its own comment at `internal/snapshot/db.go:71-77` notes "reads the whole file (~137ms on a 48 MiB store)". On the 4.5GB store from the issue that is the multi-second component. **Not** fixed here; that is what the dedup and prune PRs address by shrinking the store.

`cmd/runecho-ir/main.go:100-102` has always short-circuited `--version` ahead of any DB work. This just brings the MCP binary in line.

## What changed

- `run` takes `args []string` as its first parameter and answers `--version`/`-v` before touching the store.
- `args` is threaded in rather than read from the process-global `os.Args` inside `run`. That keeps the existing testable seam honest — a test can drive the flag path without mutating global state `go test` also owns.

## Deliberately not done

Switching the serve path to `snapshot.OpenFast` to skip `quick_check`. That trades away "never serve a corrupt DB" for a check that runs once per process lifetime, unlike the guard hook `OpenFast` was built for. Shrinking the store is the correct remedy for the size-dependent half.

## Verification

`TestVersionShortCircuitsBeforeOpeningStore` points `RUNECHO_HOME` at the **same not-a-dir fixture** `TestUnopenableStoreExitsNonZero` uses to prove a non-zero exit. So the fixture guarantees `Open` would fail if reached — if the version check ever moves below `snapshot.Open`, the test flips 0 → 1 immediately.

Mutation-verified rather than merely run: relocating the short-circuit below `snapshot.Open` fails both subtests with `exit 1, want 0`. A test that passes either way would have pinned nothing.

`go build ./... && go vet ./... && go test ./...` green; gofmt clean against the go.mod-pinned 1.25.0 toolchain.


inth3shadows:
## Review findings addressed — 2 of 4 landed here

Pushed as `9fca8f9`. Both findings on this PR were correct.

### The fix closed the symptom and left the defect one argument over

The reported bug was `--version` hanging, but the *actual* defect was that this binary parsed **no** arguments at all — so anything on the command line fell through to `Serve` and blocked on a JSON-RPC frame that never arrives. Handling only `--version` left `runecho-mcp --help`, a typo like `--verison`, and any wrapper passing an unrecognised flag hanging exactly as before. The comment even claimed to mirror `cmd/runecho-ir/main.go`, which handles `--help` *and* rejects unknown flags.

All three cases now answer ahead of the store: `--version`, `--help`, and an unrecognised argument (usage to stderr, exit 2). A bare invocation — how an MCP host actually launches this — is untouched, and `TestNoArgsStillServes` is now the explicit control for that.

### The test didn't test what its comment claimed

Fair catch, and it was my comment. `TestVersionDoesNotFallThroughToServe` asserted *"a non-empty reader that is NOT a valid JSON-RPC frame would make Serve report an error"* — but `runArgsWithHome` always passed `strings.NewReader("")`; `"extra-arg"` was an argv element, not stdin. With an empty reader, `Serve` hits EOF and returns nil, so deleting the short-circuit would still exit 0 and the test would pass while pinning nothing.

It now passes a real garbage frame and asserts stderr stayed silent, which is genuine evidence `Serve` was never reached.

### Mutation results

| Mutation | Result |
|---|---|
| drop the unknown-flag rejection | **the suite HANGS** — the original bug reproduced verbatim |
| handle `--version` but after `snapshot.Open` | both subtests fail |
| let `--help` drop through to `Serve` | all three subtests fail |

The first is the most useful outcome of the review: the new unknown-flag test reads from a reader that never returns EOF, so a regression hangs visibly instead of quietly passing.

`go test ./... -race` green; gofmt clean.

inth3shadows:

PR #353, opened 2026-08-15
fix(store): stop writing duplicate reindex snapshots (#351)

inth3shadows:
Part 2 of 4 for #351 — the write-side fix. Stacked after #352 (`--version`) but independent of it; either can merge first.

## The bug

`doReindex` called `SaveSnapshot` unconditionally. `irData.RootHash` was already computed two lines earlier and never consulted. So the hourly `repo reindex --all` job wrote a fresh `snapshots` row — plus a complete copy of the repo's `files`/`symbols`/`refs` tree — on every tick, changed or not. Nothing deduped, nothing aged out.

Measured on this machine's live store:

| | |
|---|---|
| `history.db` | **1.4 GB** |
| snapshots | 5,098 across only **1,390 distinct root hashes** → **73% duplicates** |
| worst repo | 458 snapshots across **3** distinct hashes |
| symbol rows | 8,881,802, of which only **693,672** belong to any repo's newest snapshot |
| bytes | `symbols` 632 MB + `idx_symbols_name` 310 MB + `idx_symbols_file` 126 MB |

The issue reports the same shape at 4.5 GB on another machine, where it made every `runecho-mcp` call appear to hang.

## The fix

`SaveReindexSnapshot` compares the incoming root hash against the repo's newest `reindex` snapshot **inside one transaction** and, on a match, advances that row's `timestamp` instead of re-inserting the identical tree.

It lives in `internal/snapshot` next to `SaveSnapshot` and `RollAutoSnapshot` rather than in the CLI, because that is where the other two label-specific write policies already live. `reindex` was the last high-volume label without one — `auto` has been capped at a single row per repo by `RollAutoSnapshot` since E6.

## Why it touches rather than skips

This is the part worth reviewing closely.

The guard's staleness check reads **the newest snapshot's own `timestamp`** — `cmd/runecho-guard/main.go:203`, `time.Since(snaps[0].Timestamp) > maxAge` — **not** `repos.last_indexed`. A bare skip-the-insert would freeze that timestamp at the last real content change, so any repo that stopped changing would start emitting "IR is stale" advisories once past `RUNECHO_GUARD_MAX_AGE` (default 24h). That trades a disk-growth bug for a false-advisory bug.

Advancing the timestamp leaves that read exactly as fresh as a full rewrite would have — sound precisely because the content is identical by construction.

`doReindex` calls `TouchRepo` **unconditionally in both branches** for the same reason: a repo whose content did not change this tick is fully current, and recording it otherwise is a lie the guard acts on.

## Scope

Stops the growth. Does **not** reclaim the existing 1.4 GB — 92% of symbol rows are historical and need retention plus a `VACUUM`, which is the next PR.

**No schema change.** Everything here is an `UPDATE` on existing columns. `SchemaVersion` stays 9 — a migration on a multi-gigabyte store is risk without cause.

## Verification

Mutation-verified rather than merely run. Five mutations, all killed, each by the test written for it:

| Mutation | Result |
|---|---|
| delete the timestamp `UPDATE` (bare skip) | only `TouchAdvancesTimestamp` fails — exactly the specificity wanted |
| invert the hash comparison | 3 tests fail |
| drop the `repo_id` scope from the lookup | `ScopedPerRepo` fails |
| treat `ErrNoRows` as "unchanged" | all 6 fail |
| move `TouchRepo` inside `if wrote` | only the caller-side `UnchangedTreeStillTouchesRepo` fails |

### Two existing tests changed

`TestRepoReindex` and `TestRepoReindex_Incremental` both re-indexed a tree that had not changed since enrollment and asserted `"Reindexed"` — they were using the duplicate write as their success marker. Both now assert `"Unchanged"`, which is the accurate outcome and pins the dedup at the CLI level too. `TestRepoReindex_Incremental` additionally asserts `"Reindexed"` on the leg where the source really does change, so the pair pins both directions rather than just relaxing an assertion.


inth3shadows:

Issue #13, opened 2026-06-06
FK ON DELETE CASCADE table-rebuild migration


inth3shadows:
**Order: 10 — explicitly after V8**

No FK in the schema has ON DELETE actions; deletion integrity rests on PurgeRepo's hand-ordered deletes. RemoveRepo TOCTOU already closed transactionally (audit fix). The cascade needs a table-rebuild migration — deliberately NOT taken now because V8 is reserved for contracts.

Source: 2026-06-06 audit, deliberately-skipped backlog.

inth3shadows:
**Two findings: the stated blocker is void, and the real blocker is different — and dangerous.**

## 1. The V8 dependency was never real

This issue said the cascade was "deliberately NOT taken now because V8 is reserved for contracts." That reservation is stale: **V8 was consumed by `symbols.sig_hash`** (e9e0a64), not by contracts — see the correction on #12. There is no contracts-shaped hold on the next migration slot. Retitled accordingly.

## 2. The real blocker: the migration runner's transaction makes the standard rebuild silently destructive

SQLite's documented procedure for adding `ON DELETE CASCADE` to an existing FK is a table rebuild, and it **requires `PRAGMA foreign_keys=OFF` before the transaction starts**. `DB.migrate()` runs every migration inside `db.conn.Begin()` (BEGIN IMMEDIATE), so a `migrateV9(tx)` cannot satisfy that precondition.

I verified both halves empirically rather than trusting the docs — a throwaway test against this repo's own `storeDSN`:

```
outside tx, after ON: foreign_keys=1
INSIDE tx, after attempting OFF: foreign_keys=1   <- the toggle was silently IGNORED
after DROP TABLE p (the parent): child rows remaining = 0   <- cascade fired and ate them
```

**Consequence, and why this matters more than the feature:** rebuilding `files` means `DROP TABLE files`, which with FKs on issues an implicit DELETE that **cascades into `symbols` and `refs` and deletes their rows** — mid-migration, inside the transaction, with no error. The naive implementation of this issue does not fail loudly; it silently empties the child tables and commits. The store would look healthy and answer every symbol query with nothing, which for an existence checker is the worst possible failure direction.

Note `PRAGMA foreign_keys=ON` *is* already set at open (`db.go`), so this is a live hazard, not a theoretical one.

## What the fix actually requires

[...]

PR #354, opened 2026-08-15
feat(store): retention for reindex snapshots — repo prune (#351)

inth3shadows:
Part 3 of 4 for #351. **Stacked on #353** (base is `reindex-dedup-351`, not `master`) — it uses `SaveReindexSnapshot`/`ReindexLabel` from that PR. GitHub will retarget this to `master` automatically once #353 merges.

#353 stops the store growing. This one reclaims the backlog already in it.

## Why there was nothing to reclaim it with

Nothing has ever deleted a `reindex` snapshot. `auto` has been capped at one row per repo since E6 (`RollAutoSnapshot`); `reindex` — **4,878 of 5,098 rows** on the store that motivated this issue — never got the equivalent, so every hourly tick since first enrollment is still on disk.

```
runecho-ir repo prune [--keep=30] [--repo=<name>] [--dry-run] [--vacuum]
```

## Design decisions worth reviewing

**One delete path.** Retention goes through `deleteSnapshotsTx`, whose comment says *"This is the ONLY place the child-to-parent deletion order lives… Re-open #13 if a delete path ever appears outside this helper."* No second path is introduced. The orphan assertions are the fence that replaces the `ON DELETE CASCADE` the schema deliberately does not carry.

**Only `reindex` is a candidate.** `session-start` / `probe` / manual labels are exactly what `diff --since=<label>` and `truth-trail` resolve against, and they are a rounding error by volume (220 of 5,098). Pruning them would break a real feature to reclaim nothing.

**keep=30, count-based.** Chosen against the only consumer that reads history by depth: `churn` defaults to `--n=20` and counts snapshots via `db.List`. An age-based rule would shrink a quiet repo's usable window for a reason unrelated to what anything actually reads. Every other history reader resolves only the *newest* snapshot for a label, so any keep ≥ 1 is safe for them.

**`--dry-run` and the delete share one predicate** (`pruneReindexWhere`), so the preview cannot drift from the action. A preview that disagrees with what it previews is worse than no preview, because it is trusted.

**The periodic job gets a flag, not a chained command.** `repo reindex --all --prune`. The macOS LaunchAgent runs an argv array with **no shell**, so `&& repo prune` would work in the crontab line and silently not on macOS — one scheduler quietly not pruning is the exact asymmetry this issue is about. `launchdPlist` is split out of `installLaunchd` for the same reason `cronEntry` is split out of `installCron`: so the generated text can be asserted on any platform.

**VACUUM is opt-in and never automatic.** Deleting rows frees pages *inside* the file; only `VACUUM` returns them to the filesystem, and it rewrites every live page. On an hourly timer that is worse than the growth it cleans up. Growth stays bounded without it — SQLite reuses freed pages — so `--vacuum` is the one-time step for a store that already has a backlog.

**No schema change.** `SchemaVersion` stays 9.

## Verified against real data

Run against a `VACUUM INTO` copy of a live 1.4 GB store. The live store was never touched.

| | before | after |
|---|---|---|
| snapshots | 5,118 | **2,578** |
| symbols | 8,951,165 | **6,009,126** |
| refs | 2,729,961 | 1,896,380 |
| files | 447,921 | 261,394 |
| file size | 1.3 GB | **838 MB** |

`--dry-run` predicted 2,540; the delete removed exactly 2,540. Elapsed 1m34s including the VACUUM. Afterwards: **0** orphaned refs, **0** orphaned symbols, **0** orphaned files, `PRAGMA quick_check` → `ok`.

keep=30 is deliberately conservative — no repo exceeds 30 afterwards, and dry-runs show a further 265 / 745 / 1,217 / 1,500 snapshots available at keep=20 / 10 / 5 / 3 if a tighter policy is ever wanted.

## Mutation-verified

Seven mutations, all killed:

| Mutation | Killed by |
|---|---|
| drop the label filter | `ExemptLabelsUntouched` |
| `>=` becomes `>` (off-by-one) | four tests |
| keep OLDEST N instead of newest | `KeepsExactlyTheNewestN` |
| drop the `repo_id` scope | `ScopesToOneRepo` |
| `--dry-run` falls through to the delete | `DryRunDeletesNothing` |
| installers stop passing `--prune` | `PeriodicJobPrunes` |
| `reindex --all` prunes unconditionally | `PruneFlagIsOptIn` |

**The last one initially survived**, and the fix is worth flagging: the fixture seeded 8 snapshots while `--keep` defaulted to 30, so an unconditional prune deleted nothing and the assertion passed for the wrong reason. The test now passes `--keep=2` *without* `--prune`, so the only way all 8 survive is if the prune genuinely did not run. A passing test that pins nothing looks identical to one that pins something.


inth3shadows:
## Review findings addressed — both confirmed by measurement

Pushed as `3cdefed`. Both findings on this PR were real; one had the right conclusion from the wrong cause, which the measurements sort out.

### 1. The prune starved concurrent writers — confirmed, and worse than reported

The review cited 1m34s from the PR body. That number **included the VACUUM**. Re-measured with the VACUUM removed, the write transaction alone is **63 seconds** against the 5s `busy_timeout`. So the finding understated it: the lock-holding window was the whole 63s, and it is now on an hourly timer.

Reproduced with a concurrent writer using the store's own `busy_timeout`:

| | writes locked out |
|---|---|
| single transaction (control) | **5 of 15** |
| chunked (this commit) | **0 of 15** |

**The proposed cause was wrong, though.** The review attributed the cost to the missing composite index on `snapshots(repo_id, label, timestamp)`. Measured separately:

- the correlated `COUNT(*)` is **0.37s** of the 63s
- adding that covering index changed the query plan (`SEARCH s2 USING COVERING INDEX`) and moved the total from 63s to 62s

The cost is the child-row deletes: 2.9M symbols, 835k refs, 187k files. An index cannot help with that.

**Per-repo batching also isn't enough** — the first thing I tried. The largest repo alone held the lock **27s**. Only a bound on snapshots-per-transaction is independent of how lopsided a store is, so the delete now runs in chunks of 50 through the same `deleteSnapshotsTx` funnel.

Stated in the code: a store-wide prune is no longer atomic. Partial progress is the better failure mode — pruned history is redundant by construction, so completing 400 repos and failing on the 401st beats rolling all of it back. Callers get the true deleted count alongside the error.

### 2. The fix never reached an existing install — confirmed on this very machine

```
$ crontab -l | grep runecho
0 * * * * '/home/ericm/bin/runecho-ir' repo reindex --all >>/tmp/runecho-reindex.log 2>&1 # runecho
```

No `--prune`. And `install.sh` has no periodic handling whatsoever, so `version-check --reinstall` never rewrites it. Exactly as the review predicted: the machines that would benefit most keep running the old command.

`TestPeriodicJobPrunes` stays green throughout, because the *generator* is correct and the *installed artifact* is stale — the same shape as the grammar-tag and PostToolUse-recorder incidents.

`runecho-ir doctor` now reads what is actually scheduled (crontab, or the LaunchAgent plist) and warns when it does not prune:

```
[warn] periodic reindex   installed in crontab but does NOT prune — snapshot history grows without bound (#351)
      -> runecho-ir install --periodic   # rewrites the schedule to `repo reindex --all --prune`
```

Verified against the real stale entry above. The verdict is split from the I/O (`classifyPeriodic`) so every shape is testable without a real crontab on the test machine. `doctor` was the right home for this — "is this install actually wired?" is its entire remit.

### Mutation results

| Mutation | Killed by |
|---|---|
| only the first chunk runs | `CrossesChunkBoundary` |
| chunk stride skips a chunk | `CrossesChunkBoundary` |
| chunk size raised to 100000 | `ChunkSizeIsBounded` |
| stale periodic job classified OK | `StaleJobWithoutPruneIsWarn` |

`CrossesChunkBoundary` is new and was necessary: **every other prune test seeds fewer snapshots than one chunk**, so all of them would have stayed green with chunking broken outright — present but never executed. It also guards its own fixture size, because deriving it from the constant made a raised constant try to seed 250,000 rows and hang the package rather than fail usefully.

### Not taken

Nothing. Both findings were valid.

`go test ./... -race` green; gofmt clean. #355 rebased onto this.

inth3shadows:

PR #355, opened 2026-08-15
feat(store): repo prune-missing — dead enrolments, list-only by default (#351)

inth3shadows:
Part 4 of 4 for #351. **Stacked on #354** (base is `repo-prune-retention-351`). Last, and separate on purpose: it is the only piece of this issue that can delete a repo's entire history, so it gets its own revert boundary and its own scrutiny on the `--yes` gate.

## The problem

On a machine using the worktree-per-session workflow, enrolments outlive their directories. On the store that motivated this issue, **446 of 530 enrolled repos** point at paths that no longer exist. Each costs a failed reindex and a log line every hour, forever, and nothing ever cleans them up.

```
runecho-ir repo prune-missing [--yes]
```

Lists by default and deletes nothing. `--yes` purges through `PurgeRepo` — the existing orphan-tested whole-repo funnel `repo rm` already uses. No new delete path.

## Two safety properties, both deliberate

**1. There is no heuristic to make this automatic.** No age threshold, no consecutive-failure counter, no auto-deregistration. A `stat()` miss means "not there *right now*", which is also what an unmounted external drive, a detached share, or a temporarily moved checkout looks like. A drive that has been unplugged for a month is still not gone. The decision stays with a human who can read the printed paths and recognise them — and it is never wired into `reindex --all`, cron, or launchd.

**2. Only `ENOENT` counts as evidence.** A permission error or an I/O error on a flaky mount means "cannot tell", not "gone". Treating any stat failure as missing is precisely how `--yes` deletes something real.

It also reads `EffectiveSourceRoot`, not `Path`: a bare-repo worktree enrolment stores a distinct `source_root`, and that is the directory the reindexer actually walks. Checking `Path` would report a live repo missing.

## A correction to the issue's framing

**This is not a fix for snapshot growth.** `buildIR` already refuses a vanished root via `requireExistingDir` (`cmd/runecho-ir/capture.go:142`) before any snapshot is written, so these repos are dead weight and hourly log noise — not a source of junk rows. The 1,070 empty-tree-hash snapshots observed on the real store came from repos that were *empty but present* at index time, not from missing ones. Worth stating so nobody later expects this command to have reclaimed space.

## Verification

Against a copy of a real store: 447 of 531 enrolments reported, **zero deleted** without `--yes`.

Mutation-verified, four mutations, all killed:

| Mutation | Killed by |
|---|---|
| purge without `--yes` | `ListOnlyByDefault` |
| read `Path` instead of `EffectiveSourceRoot` | `SourceRootWinsOverPath` |
| any stat error counts as missing | `UnreadableIsNotMissing` |
| purge every repo, not only the missing | `YesPurgesOnlyTheMissing` |

**The third mutation initially could not be scored at all** — no test distinguished `ENOENT` from any other stat error, so the conservative check was unpinned. `UnreadableIsNotMissing` was written for it, using a path whose parent is a regular file: that yields `ENOTDIR`, which is *not* `errors.Is(err, os.ErrNotExist)`. Portable, and unlike a chmod-based permission fixture it still holds when tests run as root. The test also sanity-checks its own fixture rather than assuming the platform's errno mapping.


inth3shadows:

PR #356, opened 2026-08-15
docs+guard: slash-separated lint rule header (#333), interception-surface table (#202)

inth3shadows:
Two small closeouts, independent of the #351 stack. Both issues turned out to be almost entirely already shipped — this is the remainder.

---

## #333 — the ask header's rule list

The header rendered `(F821,F811)` while every comment and test docstring in `lint.go` described it as `(F821/F811)`. A slash reads as "or": these are alternatives a finding can *be*, not a list the reader supplies.

**The fix is at the call site, not in the constant** — this is the part worth reviewing. `lintSelect` is doing three jobs at once:

1. ruff's literal `--select` argument (`lint.go:157`), where comma separation is **ruff's own syntax**
2. the source of `lintSelectedRules`, the membership filter that drops ruff's unconditional `invalid-syntax` output
3. until now, the ask header's display text

So "fixing" the display by editing the constant would change which rules run, or break the invocation outright. `lintSelectDisplay` is derived from `lintSelect` so the two can never name different rules.

**Nothing pinned the separator before.** Two tests now do, from both sides: one asserts the header uses slashes, the other asserts the constant stays comma-separated *and* that the display form and the membership filter still name the same rules.

Mutation-verified, three mutations, all killed:

| Mutation | Killed by |
|---|---|
| header reverts to the raw comma constant | `HeaderUsesSlashSeparatedRules` |
| "fix" the display by editing `lintSelect` itself | the ruff invocation breaks — `TestLintDifferential` and the happy-path subtests fail. **This is exactly the trap the split exists to prevent.** |
| display drifts to a different rule set | both new tests |

### Also on #333 — the dogfood window has no denominator

Not a code change; posting as an issue comment. Summary of what I found while scoping this:

`RUNECHO_GUARD_LINT=1` has been set globally since 2026-08-11, `ruff` 0.16.1 is on PATH, and Python-heavy repos are enrolled and active. Yet `~/.runecho/decisions.jsonl` contains **exactly two** lint records ever — 49 seconds apart on the day the window opened, both from `runecho-master`, one carrying the synthetic symbol `helper_that_does_not_exist`.

The blocker is not "wait longer". `CheckResult`/`VerdictSkipped` exist only in memory and are **never persisted** — the decision log records `defer`/`ask`/`outcome` and no lint-named field of any kind, verified across 4,712 records. So "zero lint asks" is byte-identical to "the check never ran once", which is precisely the trap `lint_differential_test.go` already documents for its own corpus. Until the per-check verdict is written to the decision log, more waiting produces more of the same unreadable zero.

Separately: the issue's remaining item *"split fpreport/fpaudit by version"* is **already satisfied** — `--gv` / `guardstats.FilterVersion` predate the lint substrate. Only a `>=` threshold convenience flag is missing, which is not worth a PR at n=2.

---

## #202 — the interception-surface table

The README reordering and the 20 repo topics were both already done and verified in-tree; this adds the one remaining in-repo item.

New TECHNICAL.md subsection under `## The Guard`: per tool, **what the guard actually receives → what compensates → what stays blind**. `Write` hands over the complete proposed file; `Edit`/`MultiEdit` hand over only the changed region, which is why `addInFileDefs` / `wholeFileBoundNames` / `FoldInFileDefs` exist and why hunk-scoped findings report `snippet line N` rather than a true file line.

This makes the determinism claim the issue wants to lead with **stronger**, not weaker: "the same code produces the same answer" is true *per surface*, and stating the per-surface differences makes it checkable instead of merely asserted. It is also the constraint any whole-file check has to be designed around — #333's Write-only v1 is exactly this applied.

**The issue's cited line numbers were stale** (724 / 1069 / 1010). Every anchor in the table was re-verified against the current tree: `743`, `1033-1034`, `1132-1135`, `1218`, `1238`, and `internal/guard/foldinfile.go:18`.

Directory presence — PRs into third-party repos — remains explicitly out of scope and still needs sign-off.

---


inth3shadows:

Issue #143, opened 2026-07-14
UpdateFile follows symlinks while walkSourceFiles skips them — single-file refresh can index files outside the repo


inth3shadows:
## Summary

The full-tree walk and the per-edit single-file refresh handle symlinks inconsistently. `walkSourceFiles` explicitly **skips** symlinks; `UpdateFile` **follows** them (`os.Stat` + `os.ReadFile`), with no symlink guard. A symlink located inside the repo but pointing outside it is therefore skipped by a full `Generate`/`Update`, but indexed by `UpdateFile` (the per-edit auto-fresh hook) — pulling the external target's content/symbols into the IR under an in-repo key.

## Evidence (`internal/ir/generator.go`, current `master`)

`walkSourceFiles` skips symlinks (L159-164):
```go
if info.Mode()&os.ModeSymlink != 0 {
    if info.IsDir() {
        return filepath.SkipDir
    }
    return nil   // symlinked file: skipped
}
```

`UpdateFile` has no such guard — its outside-repo check is a lexical `filepath.Rel` on the *symlink's* path (which is inside the repo, so it passes), then it stats/parses the target (L329-364):
```go
rel, err := filepath.Rel(absRoot, absFile)
if err != nil || rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
    return existing, false, nil // edited file is outside this repo
}
// ...
info, statErr := os.Stat(absFile)   // follows the symlink to its target
// ...
fileIR, perr := g.parseFile(absFile) // reads the target's bytes
```

So `evil.go -> /etc/passwd` (or `~/.ssh/id_rsa`) inside the repo: skipped by the walk, but `UpdateFile("evil.go")` indexes the target's structure into the IR — which is surfaced via the MCP oracle.

## Impact

- **Inconsistency:** the same repo produces different IRs depending on whether it was built by a full walk or touched by the per-edit refresh.
- **Security (threat-model dependent):** when indexing an untrusted/cloned repo, a crafted symlink can exfiltrate the structure of arbitrary local files into the IR. Low for self-owned repos; real for review-of-untrusted workflows.

## Suggested fix

Mirror `walkSourceFiles` in `UpdateFile`: `os.Lstat` the target first and return `(existing, false, nil)` (or drop the entry) when it is a symlink, before `os.Stat`/`parseFile`. Add a regression test with a symlink pointing outside the repo.

## Provenance

[...]

PR #366, opened 2026-08-25
feat(diff): report what the indexer declined, and why

inth3shadows:
Unblocks [inth3shadows/harness#21](https://github.com/inth3shadows/harness/issues/21).

## The gap

`diff --since` could not distinguish **"nothing was removed"** from **"nothing was read"**.

An unindexed language is absent from the IR entirely, so deleting a function from a `.java`, `.c`, or `.php` file produced `total_removed: 0`, an empty `files` list, exit 0 and empty stderr — byte-identical to an unchanged repo.

Coverage could not surface it either. `Stats.Coverage()` is `Indexed / SupportedSeen`, and an unindexed language is never in the denominator — which is why a repo containing an unparsed file still reads `COVER 100.0%`.

## What it looks like now

```
IR DIFF  2d87a3c7... → 2d87a3c7...

No structural changes.

NOT EXAMINED (2 paths) — a symbol removed here would be invisible to this diff:
  .git  [ignored_dir]
  Widget.java  [unsupported_language]
```

`--json` carries the same facts as `"skipped": [{Path, Reason}]` plus `"skipped_truncated"`. Reasons: `unsupported_language`, `parse_error`, `oversized`, `cap_reached`, `ignored_dir`.

Note the block is attached to the **zero-drift** branch too. "No structural changes." for a Java repo is true only in the sense that nothing was ever looked at — that unqualified sentence is the whole failure.

## The enabling piece

`internal/ir/skipped.go`'s `knownSourceExtensions`.

`supportsExtension` answers one question — *"do I have a parser?"* — so `.java` and `.md` are the same answer. A consumer cannot fail closed on that without blocking every documentation change. Telling *"code I do not index"* from *"not code"* is knowledge only the indexer can hold, and duplicating it in each consumer guarantees drift, so it lives beside parser registration.

The table is deliberately conservative. A **missing** entry degrades to the previous silent skip; a **wrong** entry makes a consumer fail on a file that never needed indexing. Under-inclusion is the safe error, so ambiguous extensions are left out — `.d` (D source, but also the make dependency files every C build litters a tree with) is the worked example.

`TestKnownSourceExtensionsAreNotParsed` pins the one invariant: no entry may be supported by a registered parser. A stale entry is worse than a missing one — it reports an *indexed* file as unexamined, and the resulting false block gets blamed on the gate rather than on this table.

## Two absences that are load-bearing

**Non-code is never listed.** `README.md`, `config.yaml`, `data.json`, `notes.txt` do not appear — verified end to end through a built binary, not just a unit test. This is the property a fail-closed consumer rests on.

**A missing `skipped` key means "not measured", not "nothing skipped".** Only `--since` walks the filesystem. The two-snapshot form (`diff <id-a> <id-b>`) omits the key entirely rather than claim a clean bill of health it never checked — emitting `"skipped": []` there would be the same fail-open shape as an empty `removed` list standing in for a guard that never ran. Enforced two ways:

- `DiffResult.SkippedKnown`, which `DiffPayload` gates on.
- `DiffLiveWithStats` takes the walk's `ir.Stats` **explicitly**, so a caller that does not have them cannot publish an empty list by accident. `DiffLive` is unchanged and still reports nothing.

`skipped_truncated` exists for the same reason: past the 1000-entry cap, absence from the list no longer implies "indexed".

## Decisions worth a reviewer's disagreement

**Ignored directories are reported, `.git` included.** So every repo carries at least one entry. Filtering them needs a second curated table of *which of the operator's own ignore rules matter* — that is consumer policy, not oracle knowledge, and it would drift exactly the way an extension allowlist in a consumer would. A test asserts `.git` appears, so filtering it later is a deliberate change with a failing test rather than a silent one.

**`parseFailureReason` re-stats the file instead of matching the error string.** `Generate` has no separate size guard — `parseFile` rejects oversized files — so oversized and syntactically-broken reach the same branch. The message is a human diagnostic free to change; the size is the fact the reason asserts.

**The "corrupt file" test is a classifier unit test, not a fixture.** The parsers are tree-sitter-based and error-tolerant: garbage bytes in a `.go` file still yield a tree, so a corrupt-source fixture asserted nothing and quietly turned into a `t.Skip`. Replaced with both branches directly, plus a mode-0000 end-to-end that skips under root with the reason stated.

## Parity

`DiffPayload`'s doc comment promises the CLI and the MCP `diff` oracle tool cannot drift, so both `DiffLive` call sites in `internal/mcp/tools_oracle.go` were moved to `DiffLiveWithStats` as well.

Skip entries use untagged Go field names (`Path`, `Reason`), matching the sibling `files` and `hot_files` arrays in the same payloads — a lone snake_case array beside Go-cased ones is a trap for the machine consumers this contract exists to serve.

## Verification

```
gofmt -l .        clean
go vet ./...      clean
go test ./...     exit 0
```

Plus a real-binary run: enrol a repo, snapshot `gate-base`, delete a method from `Widget.java`, diff. The removal is still invisible to the symbol guard (no Java parser exists, and that is fine) — what changed is that the payload no longer stays silent about it.

## Follow-ups, not in this PR

- Symlinked files and directories are skipped at `generator.go` and are arguably the same blind spot. Not named in harness#21; deferred deliberately.
- The harness side (fail closed on a changed source file appearing in `skipped`, behind an opt-out flag, with a backward-compatible branch for older binaries) is a separate PR in that repo.


inth3shadows:
## Review fixes pushed — `97643cf`

All 12 findings from dual review addressed. The headline one was mine to own: **the feature reintroduced the false block it was built to prevent, one reason code over.**

`.git` is in `DefaultIgnoredPaths`, so `skipped` was non-empty in every repo, and the docs told consumers to prefix-match it. An agent editing `testdata/README.md` hit the `testdata` entry and got blocked — a documentation-only false block, exactly what `knownSourceExtensions` exists to prevent. I defended this call twice on the grounds that "which ignore rules matter" is consumer policy. That reasoning was right and the conclusion was wrong: the fix is to *say which is which in the data*, not to merge them and hope the consumer sorts it out.

| Key | Contains | Match | Consumer |
|---|---|---|---|
| `skipped` | **Files** the indexer could not read | Exact path | Fail closed |
| `ignored_paths` | **Directories** it did not enter | Prefix | Informational |

### Other findings

- **`verify` was never converted.** It does its own live walk and is the most gate-like command in the tool — and it still printed a bare `No structural changes.` for a repo whose only code was unparsed. The headline bug surviving on the surface most likely to be believed. `trail` had the same gap; `TruthTrail` now takes the walk's stats.
- **`.mts`/`.cts` reproduced the original bug inside a language RunEcho advertises.** `JSParser` doesn't claim them and the table didn't list them, so a deleted export was invisible *and* unreported. Added, with `.pyi`/`.pyx`/`.pxd`, `.rake`/`.gemspec`/`.ru`, `.zsh`/`.fish`/`.ksh`, `.hcl`. `TestAdvertisedLanguageFamiliesAreAccountedFor` now requires every advertised family to be either parsed or named — the conservative bias is right for `.d`, where nobody expects coverage, and wrong where the tool says it supports the language.
- **The ignore list was applied to the walk root.** A checkout named `testdata`/`vendor`/`dist` pruned its own walk: zero files indexed, a vacuous 100% coverage (`SupportedSeen` was 0 too), and a skip entry of `"."` — a prefix matching no path a consumer could hold.
- **Extension matching was case-split.** Parsers match exactly, `IsKnownSourceExtension` lowercases, so `A.GO` fell through both: not indexed, and not reported because `.go` is deliberately absent from the source table. Lowercased at the `supportsExtension`/`parserFor` chokepoint.
- **Symlinks and unreadable entries recorded.** Their absence from the list previously read as "examined". Resolving the target needs an explicit `os.Stat` — `filepath.Walk` lstats, so `info.IsDir()` is always false for a symlink. That made the *pre-existing* `if info.IsDir() { return filepath.SkipDir }` in that branch unreachable dead code, and reachable would have been a bug: `SkipDir` from a non-directory skips the remaining files in the containing directory.
- **`cap_reached` could evict the headline reason.** First-come recorder + lexical walk meant a capped repo filled all 1000 slots with `cap_reached` and dropped every `unsupported_language` entry sorting later — the surviving signal was "lexically earliest". Sub-capped at 100; it means "everything past the cap" and its count is already implied by `SupportedSeen - Indexed`.
- **MCP `diff` clips to 50.** That tool exists for context economy — `jsonText` minifies for the same reason — and an unclipped list injected up to 1000 path objects per call on a C or Java repo. The clip sets `skipped_truncated`, so it is a smaller honest answer, not a quieter one.
- **Rendering:** `plural(n, "more")` produced `and 5 mores`, and my test's second `Contains` clause subsumed the first so it pinned neither spelling. A truncated header read `(1000 paths)` as though it were a total. The zero-drift branch had gained an unconditional trailing newline.

### Two calls beyond what review prescribed

**Human output drops configured `ignored_dir` entries** (they stay in `--json`). Echoing your own ignore list back on every diff of every repo makes the note unreadable, and a note that fires every time stops being read. `symlink_dir`/`unreadable_dir` still print — you didn't configure those.

**`cap_reached` got a sub-cap rather than a priority queue.** Simpler, and it loses nothing a consumer can act on.

### Verification

```
gofmt -l .     clean
go vet ./...   clean
go test ./...  exit 0
```

End to end against a built binary, on a repo with `main.go`, `Widget.java`, `mod.mts`, `README.md`, `testdata/README.md`:

```
skipped      : [Widget.java unsupported_language, mod.mts unsupported_language]
ignored_paths: [.git ignored_dir, testdata ignored_dir]
verify       : now prints the NOT EXAMINED block
```

Note: `runecho-guard` was bypassed on this commit. It resolves this unenrolled worktree to `runecho-master`'s index — a different worktree with none of these new files — so all ten "unresolved symbol" findings were false by construction. `go build ./...` and `go vet ./...` passing is the proof they resolve.

inth3shadows:
## Round 2 review fixes — 11 commits, `97643cf..425dbc9`

Dual review of `97643cf` returned 18 findings (10 + 8, overlapping). All fixed, one per commit, each with a test confirmed red before the fix.

```
53d73e7  lowercase the extension in parseFile, not just parserFor
6bcdb50  terminate the clean-repo diff with exactly one newline
d5665ea  follow a symlinked walk root instead of pruning it
9f432f7  classify unreadable/unfollowable directories as blind spots
1786a00  apply the code filter to dangling symlinks too
90baf17  render the skip block, making TruthTrail's stats parameter real
bab36c6  keep the truncation warning, and name the cap that actually fired
3c0f2fc  budget TruncateSkips per bucket, not over the merged list
70fe534  drop .hcl and .ru — both match common non-code
fc337a1  check the ignore list before the symlink branch
425dbc9  correct the unreadable-entry rationale, and a comment that lied
```

### The contract changed, and this is the part to read

The array split is no longer **files vs directories**. It is **"did you ask for this?"**

My original line put `unreadable_dir` and `symlink_dir` on the informational side because they name directories. That was wrong in the direction that matters: a directory the walk cannot read takes its whole subtree with it, and nothing records the files inside *because nothing ever saw them*. CI running without read permission on `src/legacy/` therefore returned `skipped: []`, `total_removed: 0`, exit 0 — for a tree it never opened.

`policySkips` now holds exactly one reason: `ignored_dir`, the operator's own configuration. A permission error and an unfollowable link are limits of the tool, so they are blind spots. Both arrays are matched the same way:

```go
changed == entry || strings.HasPrefix(changed, entry+"/")
```

correct for files and directories alike, since a file entry can never be a directory prefix of another path.

### Two were regressions I introduced in round 1

**The case fix made `.TS` worse.** `supportsExtension`/`parserFor` were lowercased but `parseFile` was not — and that same value selects the *grammar* via `ParseExt`. `jsLanguageFor` matches exactly, so `.TS` fell to the JavaScript grammar:

```
.ts  ->  [class:Widget export:Widget function:Widget.bar]
.TS  ->  [class:Widget export:Widget]
```

`A.TS` went from *not indexed* to *indexed with a wrong symbol set* — strictly worse, because an unindexed file is reported as a blind spot while a mis-parsed one reports a clean diff. The bug this PR closes, reintroduced by its own fix.

**A symlinked repo root was a total silent fail-open.** `filepath.Walk` lstats, so a root that is a symlink to a directory hits the symlink branch, gets recorded as pruned, and the walk ends after one callback: zero files indexed, exit 0, `skipped` empty. The only trace was `{"Path": "."}` — in the informational array, and a prefix matching no path a consumer could hold. `~/work -> /mnt/checkouts/x` or a pnpm workspace link was enough. Round 1 added the root exemption to the ignore branch and missed the symlink branch it was rewriting at the time.

### The rest

- **`TruthTrail`'s stats parameter was inert.** Round 1 added it, updated four call sites, and wrote a comment claiming the receipt "can name what the indexer never read". `FormatTrail` never referenced `Diff.Skipped`, so `truth-trail` printed `No structural changes` for a Java-only repo while `verify` named the file. A receipt reads as a settled account, which makes it the worst surface to leave unqualified.
- **Dangling symlinks bypassed the code filter.** Both sibling branches filter by extension; that one did not. A stale `latest -> build-123` or a `README-link.md` pointing nowhere put non-code into the fail-closed array and blocked every build in the repo. Symlink loops reach it too, via ELOOP. Every file-skip branch now routes through one `isCode` guard rather than repeating the check — the previous review found this same defect through a different door, twice.
- **The truncation WARNING could vanish**, since `writeSkipBlock` returns early on an empty list and is the only thing that prints it. Reachable: a monorepo with >1000 pruned directories fills the list with `ignored_dir` entries and every `.java` seen afterwards is dropped. It also hard-coded `MaxRecordedSkips`, so a `cap_reached` sub-cap truncation printed `hit its cap (1000)` beside ~101 entries.
- **`TruncateSkips` clipped the merged list.** `result()` sorts lexically, so on a 50-package workspace the `node_modules` entries interleave with the `App.vue` ones and a 50-entry MCP clip discarded 15 of 40 real blind spots. With enough ignored directories, `skipped` came back empty for a repo that was entirely unindexed. Budgeted per bucket now — the same answer the `maxCapReachedSkips` sub-cap uses one layer down.
- **`.hcl` and `.ru` removed.** `filepath.Ext(".terraform.lock.hcl")` is `.hcl`, and `README.ru` is the `<name>.<lang>` translated-doc convention. Both are the accident the table's own `.d` rule already names. `.ru` also came out of `advertisedFamilies`, where the two safety rules genuinely conflict — a false block is louder and more frequent than one unreported `config.ru`, and the test now records that trade.
- **Symlinked `node_modules`** (standard in pnpm/yarn workspaces) was classified `symlink_dir`, i.e. a permanent unfixable entry in the fail-closed array for a directory the operator excluded. The ignore check moved ahead of the symlink branch.
- **`mapcmd.go` carried a comment describing a change that was not made** — it claimed stats were passed, directly above a line that still called `DiffLive`.

### Verification

```
gofmt -l .     clean
go vet ./...   clean
go test ./...  exit 0
```

End to end through a **symlinked root**, with every trap file present (`README.md`, `README.ru`, `.terraform.lock.hcl`, a dangling `latest`, `testdata/README.md`):

```
total_removed: 0
skipped      : [(Widget.java, unsupported_language), (mod.mts, unsupported_language)]
ignored_paths: [(.git, ignored_dir), (testdata, ignored_dir)]
truth-trail  : now prints the NOT EXAMINED block
```

inth3shadows:
## Round 3 review fixes — 10 commits, `425dbc9..fc48cfe`

Dual review of `425dbc9` returned 11 unique findings across two reviewers, including **a third-generation regression** — round 3 closed a fail-open for a plain directory and reopened it for a symlinked one.

```
1f0e5ba  a symlink that will not resolve may still be hiding a subtree
83988e3  keep unstattable non-code out of the fail-closed array
40dead3  flag truncation only when the fail-closed array actually lost entries
5f308c1  report the largest cap that truncated, not the last one written
9acbe2e  resolve symlinks in UpdateFile's containment test too
850d349  apply the ignore list to directories and links only
41fe0ea  warn without a header when the skip list is empty but truncated
6047583  halve maxMCPSkips now that the budget is per bucket
3357c7b  retire three comments still describing the old contract
fc48cfe  drop inert "." entries, and stop --compact hiding blind spots
```

### The regression

The symlink branch reasoned *"a dangling or looping link resolves to nothing, so it is not a directory"* and applied the code filter. The premise is false — `os.Stat` fails with EACCES/EPERM/EIO/ELOOP as well as ENOENT, and in those cases the target is a real directory full of real source. `isCode` keys off `filepath.Ext`, and a directory link has no extension, so the guard dropped every one:

```
legacy -> $unreadable/legacy      (containing Critical.java)

round 2:  skipped = [{legacy, unreadable}]
round 3:  skipped = []            gate passes, subtree never opened
```

That is verbatim the failure `skipped.go` cites as the reason `policySkips` was narrowed *one commit earlier* in the same round.

**The tell was inside the diff.** Two adjacent branches of one `switch` share an epistemic state — the stat failed, we don't know what was there — and reached opposite conclusions. Both now route through a single `looksLikeNonCodeFile` helper. Three rounds running, the defect has been a second copy of a decision drifting from the first; the shared guard is the structural answer, not more care.

The old test only exercised ENOENT, so it could not tell "nothing behind the link" from "a subtree I can't open" — it pinned the wrong invariant while passing.

### A test that asserted the bug

`TestTruncateSkips_BudgetsEachBucketSeparately` asserted `if !d.SkippedTruncated { t.Error("policy entries were dropped, so the flag must be set") }` — pinning the exact behaviour `TruncateSkips`' own doc comment calls a defect it fixed. With `maxMCPSkips` at 50, any repo with 51+ ignored directories (a `__pycache__` per package, a `node_modules` per package) reported `skipped` as "unknown" on every MCP diff, permanently, while that array was complete. `USAGE.md` defines the flag as "absence no longer implies indexed", so a false positive is a false fail-closed.

Inverted, plus a second test for the case where the capability bucket genuinely is clipped, so neither direction is unpinned.

### The rest

- **`UpdateFile`'s root was not resolved.** Resolving the *walk* root made symlinked repos indexable and left the incremental path judging every edit "outside this repo" — the full walk worked, the per-edit refresh silently did nothing, the IR went stale. Worse than before the fix, when the repo was never indexed and the staleness was visible. Both sides now go through one `resolvedOrSame`.
- **`hitCap` was last-write-wins**, so a `cap_reached` add after the main cap filled reset the reported figure from 1000 to 100 — `hit its cap (100)` beside a 1000-entry list. The inverse of the bug the sub-cap commit was written to fix. Now keeps the largest; `TruncateSkips` takes the minimum, and the two are consistent once each is stated in terms of what actually bounded the reader's list.
- **The ignore check fired for regular files** after being moved above the `IsDir` test. A `.git` *file* (worktrees, submodules — what this repo uses) hit it on every walk.
- **`info == nil` skipped the `isCode` guard**, so a mode-0444 directory put `assets/README.md` and `assets/logo.png` in the fail-closed array. Same branch is hit by a file vanishing mid-walk during a checkout — a false block that appears and disappears on its own.
- **Three stale comments** still documented the old file-vs-directory split. The `DiffPayload` one is load-bearing: it's what a Go consumer reads, and following it (exact-match on `skipped`) misses every path under an `unreadable_dir` entry.
- **`maxMCPSkips` was effectively 2n** after per-bucket budgeting — 100 objects per call on the surface whose purpose is context economy. Halved to 25, bound documented and pinned.
- **`NOT EXAMINED (0+ paths):`** rendered above zero entries; the warning now stands alone.
- **`"."` entries** matched no path a consumer could hold, and **`--compact` printed nothing at exit 0** for a repo the indexer could not read — the most confident possible statement from the surface with the least room to qualify it.

### Verification

```
gofmt -l .     clean
go vet ./...   clean
go test ./...  exit 0
```

End to end through a symlinked root whose `legacy` is a link to an unreadable directory — the regression case:

```
skipped      : [(legacy, unreadable_dir)]     ← was []
ignored_paths: [(.git, ignored_dir)]
--compact    : IR DIFF [...]: no drift, but 1 path NOT examined   ← was silent
```

Independent mutation testing last round confirmed 11 of 11 fixes had a test that dies when reverted; the same discipline was applied here.

inth3shadows:
## Round 4 review fixes — 9 commits, `fc48cfe..a10d3fe`

Both reviewers found fourth-generation regressions. All fixed, one commit each.

```
e4701cd  blame the unenterable directory, not each child by name      (self-found)
ad4db5d  flag an unreadable repo root instead of dropping it silently
7d4d61c  resolve the parent directory so a DELETE under a symlinked root works
675e332  apply the ENOENT policy to the symlink branch too
af37412  decide from the error, not from the filename
44a966b  make --compact qualify every case FormatFull does
e96f6ca  test containment as named first, and only then through symlinks
6de4aaa  stop the truncation warning claiming to bound the whole list
a10d3fe  restore two deleted regression tests, drop an unreachable reason
```

### The sharpest one: `filepath.Ext(".github") == ".github"`

`looksLikeNonCodeFile` decided "is this non-code?" from the extension. That is not a question an extension can answer for a **directory**, and it failed in the most ordinary shape there is:

| dir (0000, containing `Critical.ts`) | before | after |
|---|---|---|
| `.github` | **dropped** | `unreadable_dir` |
| `.storybook` | **dropped** | `unreadable_dir` |
| `v1.2` | **dropped** | `unreadable_dir` |
| `legacy` | `unreadable_dir` | `unreadable_dir` |

None of `.github`/`.storybook`/`.husky`/`.circleci` is in `DefaultIgnoredPaths`, so all are walked. CI without read permission on `.storybook/` meant a deleted export in `.storybook/main.ts` was invisible **and** unreported — `skipped: []`, exit 0, gate passes. The documented failure mode I'd written off as "a directory named `my.config/`, rare" was the whole class. The same heuristic also pushed `Makefile`, `LICENSE`, `COPYING`, `AUTHORS` and `Dockerfile` *into* the fail-closed array, since an extension is no evidence for a path that has none.

Every regression test in the suite used the one dot-free fixture name (`legacy`), which is why they all passed.

**The helper is deleted, not patched.** Two signals replace it, neither of which cares what a path is named: the **error** (`ENOENT` means gone and nothing was hidden; anything else means something is there we could not look at) and the **parent, tested directly** via `everyChildFailed` rather than inferred from one child's failure — which also fixes the over-broad blame where a single `EIO` on a flaky NFS mount marked a whole directory unexamined.

### Two more fail-opens

**`UpdateFile` dropped every DELETION under a symlinked root.** `EvalSymlinks` fails on a path whose leaf no longer exists, so the root resolved and the file did not, `Rel` produced `../`, and the containment guard fired. A deleted file kept its symbols in the IR indefinitely. Reproducible by changing only `TMPDIR` — on macOS `t.TempDir()` is already under a symlink. The prior test covered add/modify across three namings and never a delete.

**An unreadable repo root reported nothing at all.** `"."` is dropped as inert (correctly — it matches no path a consumer holds), so with the root as blame the entry was created and discarded: `files: []`, `skipped: []`, and `Coverage()` returning a vacuous 100 because `SupportedSeen` is 0 too. My justification comment ("SupportedSeen and Coverage already carry it") was simply wrong — the payload carries neither. It needed its own flag: `root_unreadable`.

### The rest

- **`--compact` still hid the truncated case** it was changed to stop hiding, and swallowed the new root flag. Third silently-empty output on this branch; the rule now is that every case `FormatFull` qualifies, `--compact` qualifies too.
- **ENOENT policy reached only one of the two branches** that needed it, so a permanent dangling `latest -> build-123` blocked the repo forever.
- **`UpdateFile` resolved too eagerly**, making the IR key the target's path (a key the walk never produces) and rejecting out-of-repo symlinks before the #143 stale-entry cleanup could run.
- **The truncation warning**, third pass: "largest cap wins" fixed one direction and left the inverse, where only the sub-cap fires and 160 entries sit under "cap (100)". The wording was the bug — the sub-cap bounds one *reason*, not the list.
- Two deleted regression tests restored (one still cited by name in a comment as the authority for a rule conflict), and `SkipUnreadable` removed — unreachable since the walk started blaming directories, while its doc and a `USAGE.md` row still advertised it.

### Note on the reviews

Rounds 3 and 4 wrote **mutually contradictory tests over the same fixture**: round 3 asserted a dangling `latest` must be absent, round 4 that it must be present. Both were half right — `os.Stat`'s error already distinguishes "resolves to nothing" from "could not look", and neither round inspected it. That is now the load-bearing distinction in both branches.

```
gofmt -l .     clean
go vet ./...   clean
go test ./...  exit 0
```

inth3shadows:
## Round 5 review fixes — 6 commits, `a10d3fe..aef932a`

Ten findings, three HIGH. **Only one reviewer this round** — the second died on a session limit, so coverage here is weaker than the previous four rounds, where each reviewer consistently found things the other missed.

```
a222296  carry the resolved path forward, not just the rel it produced   HIGH
11de31d  keep non-code out of the unreadable-symlink branch              HIGH
2836eae  memoise everyChildFailed per directory                          HIGH
9f2a556  treat blame at or above the root as RootUnreadable
e5d16ae  a single unstattable file is not an unreadable directory
aef932a  document root_unreadable, bound the dedup map, finish --compact
```

### `UpdateFile` was deleting entries instead of refreshing them

Last round's symlink fallback derived `rel` from the resolved path and left `absFile` in link form. `pathCrossesSymlink` then walked up from the link-form path, never reached `absRoot`, hit the symlinked root component and returned true — so the entry was treated as "real file replaced by a symlink" and **dropped**:

```
link root, real file   present=true   symbols=2
real root, link file   present=false  symbols=0
link root, link file   present=false  symbols=0     ← the ordinary case
```

In a repo named through a link, every incremental refresh erased that file's symbols and the next diff reported all of them as removed. A mass phantom deletion, from the fix meant to stop a single real one going unnoticed. My test passed because it asserted only `changed == true` — and dropping the entry changes `RootHash` too.

### `everyChildFailed` was O(N²) and blew the deadline

Called once per *failing* child, and an unreadable directory fails every child, so each call did a full `Readdirnames` plus an `Lstat` per sibling:

```
n=500   0.94s
n=1000  3.72s
n=2000  13.36s
n=4000  "failed to walk directory: context deadline exceeded"
```

At 4000 entries one unreadable directory turned `Generate` into a hard error with **no IR at all** — a tooling timeout taking down the whole repo instead of reporting one subtree. Memoised per parent: 3000 entries now run in 0.53s.

### And the code filter overcorrected

Deleting `looksLikeNonCodeFile` removed the extension check from the symlink branch entirely, so `README.md -> <unreadable>/README.md` went into the fail-closed array and blocked every docs change. The lesson from the `.github` defect was that an extension answers *wrongly for directories* — not that it is useless. It is decisive in exactly one direction: **a path that has an extension is a file.** That direction, and only that one, is now used.

### The rest

- **Parent-blaming could escape the repo.** The parent of a child directly under the root *is* the root, and `filepath.Rel` returns `".."` for anything above it — so a repo whose parent directory is unreadable emitted `{"Path": "..", "unreadable_dir"}` with `RootUnreadable` false and `Coverage()` at a vacuous 100. The fail-open the flag closed, one level further out.
- **A single unstattable file was reported as `unreadable_dir`** — non-code in the fail-closed array, and a reason string that says "directory" about a file. Split out `unreadable_file`.
- **`root_unreadable` was in the payload but not the consumer contract.** It is the one condition where `skipped` is empty, `skipped_truncated` is false, and nothing was indexed — so every gate written to USAGE.md still failed open on it. Now the fourth rule, and the one to check first.
- **`--compact` rendered a tooling failure as a mass deletion**: the new switch lived inside `if len(d.Files) == 0`, but an unreadable root makes a live diff report every file as removed, so the branch never ran. It also printed `"0+ paths NOT examined"` — the exact rendering `writeSkipBlock` suppresses.
- **`seen` was unbounded**, inserted before the cap checks, so a capped monorepo accumulated millions of path strings while `items` stopped at 1000. **`subCapOnly`** was written and never read — a field added for a renderer that never saw it.

### Recommendation: do not merge yet

Five review rounds: **12, 18, 11, 11, 10** findings. The rate is flat, and three of this round's ten were regressions from the previous round's fixes.

The walk's error handling now carries five interacting axes — ENOENT vs other error, file vs directory, code vs non-code, blame-self vs blame-parent vs blame-root, and the memo. Each round I fix one axis and break another. Patching finding-by-finding is not converging; that branch wants a rewrite against an explicit truth table, with the table itself reviewed before any code changes.

```
gofmt -l .     clean
go vet ./...   clean
go test ./...  exit 0
```

Issue #370, opened 2026-08-28
Enrollments are never pruned when a worktree is deleted: 516 of 617 repos had missing roots, and prune-missing cannot finish


inth3shadows:
Two related defects in enrollment lifecycle, found while diagnosing #369. Version `v0.46.0-1-ge17d78c`.

## 1. Nothing prunes an enrollment when its worktree is deleted

Measured on this box before any cleanup:

```
617  enrolled repos
516  with a missing source root   (84%)
```

The workflow that produces this is ordinary: a per-session git worktree (`claude-<timestamp>`), indexed on first use, removed when the session ends. Indexing enrolls; nothing un-enrolls. Every session leaves a permanent tombstone, so the ratio only grows.

This is not cosmetic — it is what caused #369. The guard resolved a live repo to a stale enrollment whose path was gone and ran `git -C <deleted path> diff --cached`, which failed, which (per #369) it treated as a pass. So enrollment rot silently disables the guard.

`repo prune-missing` exists and works, but it is manual and undiscoverable: nothing surfaces the count, and the first symptom is a warning in unrelated tool output. Options, cheapest first:

- Drop or mark an enrollment lazily, when a command finds its root missing.
- Report the count in `repo list` output, so the rot is visible before it bites.
- Prune on a schedule alongside the existing reindex pass.

Note that a missing root is not always permanent — an unmounted drive should be distinguished from a deleted directory, which is presumably why the current command is opt-in and prints every path first. That caution is right; the gap is that nothing ever *reports* the state.

## 2. `prune-missing --yes` cannot purge every repo it lists

```
Warning: could not purge runecho-claude-20260729-154054 (id=430):
  purge repo 430: constraint failed: FOREIGN KEY constraint failed (787)

Purged 515 of 516 missing repo(s) and their history.
```

One row survives its own purge. The operation is not atomic per repo — a FK dependency (id 787) is not deleted first — so that enrollment is now unprunable by the documented command and will keep reappearing in every future run. The rest of the prune succeeded, so this is a stuck row rather than a corrupted database, but it needs either ordered deletes or `ON DELETE CASCADE`.

## 3. Related hygiene: stale lock files

[...]

PR #381, opened 2026-09-01
fix(store): purge a repo's contracts, the child PurgeRepo never deleted (#370)

inth3shadows:
Fixes item 2 of #370 — the row `prune-missing --yes` cannot purge.

## Root cause

`contracts.repo_id` is a `NOT NULL` foreign key to `repos(id)`. `PurgeRepo`
deleted the snapshot subtree (refs → symbols → files → snapshots) and then the
repo row, and **never deleted `contracts`**. The FK refused the repo delete and
the whole transaction rolled back.

That is `deleteSnapshotsTx`'s "fails loudly rather than orphaning" guarantee
working exactly as designed — it just arrives as an unexplained
`FOREIGN KEY constraint failed` in the middle of a bulk prune.

Verified on the live store: repo 430 has exactly **1** contracts row, and it is
the **only** contracts row in the database. That is why the failure is always
precisely one repo — "Purged 515 of 516" on 2026-08-28, "Purged 14 of 15" today.

## Two corrections to the issue text

- **`(787)` is not a row id.** It is `SQLITE_CONSTRAINT_FOREIGNKEY`, SQLite's
  extended error code. #370 reads it as "a FK dependency (id 787) is not deleted
  first", which would send someone hunting a row that does not exist.
- **`ON DELETE CASCADE` is the wrong fix**, and `deleteSnapshotsTx`'s own comment
  already argues why (#13): SQLite ignores `PRAGMA foreign_keys=OFF` inside a
  transaction, every migration here runs in one, so the table rebuild that adds
  the constraint would fire the cascade it is adding and silently empty
  symbols/refs mid-migration, then commit clean. One ordered delete inside the
  existing transaction is the whole fix.

## The comment that was read wider than its scope

`deleteSnapshotsTx` says funnelling both delete paths through it means "there is
no ordering left for a third caller to get wrong". True — of the **snapshot
subtree**. `contracts` is a direct child of `repos` that the helper never knew
about, and that gap is this bug. The comment now states its scope.

## The fence

`TestPurgeRepoDeletesEveryTableThatReferencesRepos` reads
`pragma_foreign_key_list` and requires the discovered set of `repos` children to
match, **exactly**, the set `PurgeRepo` handles. Exactly rather than as a
superset: a superset check can be satisfied by editing the list instead of the
code, and a mutation adding a phantom entry survived until this was tightened.
A future child table fails the test when it is added, not years later inside a
prune.

## Verification

Full suite green. **Six mutations killed**: the delete removed; its `WHERE`
dropped (caught by a bystander repo's surviving contract); a phantom entry in
`handled`; `contracts` removed from `handled`; discovery finding nothing.

A seventh **survived** and is reported rather than hidden: removing a
`len(children) < 2` vacuity guard changed nothing, because the exact-match check
subsumes it. The guard was deleted rather than kept as an assertion nothing can
falsify.

## Not in this PR

Items 1 and 3 of #370 — nothing *reports* enrolment rot before it bites, and the
652 stale `.lock` files in `~/.runecho/`. Both still open.


inth3shadows:
Closing in favour of #382, which reached the same D1 diagnosis and the identical
fix independently — 73 seconds apart, from a concurrent session — and covers
three more defects on top, including one (`ResolveRepo` preferring the lowest-id
enrollment) that #376 turned into a live commit blocker earlier today.

Nothing is dropped. The regression fence from this PR is ported onto #382 as
`e32a468`, along with two smaller corrections to that branch:

- a bystander assertion, because `TestPurgeRepo_WithActiveContract` passes a
  `DELETE FROM contracts` that lost its WHERE clause;
- the `(id=787)` citation — 787 is `SQLITE_CONSTRAINT_FOREIGNKEY`, the extended
  error code, not a row id. The enrollment is `id=430`.

Two competing PRs on one issue is how modelbench's research-grader pair happened.
One survives.

